PR #2, opened 2025-01-18
Fix/secrets tracking

evalstate:
Use `.secrets.yaml.example` files rather than `.secrets.yaml` to avoid user footguns.

### DIFF
--- a/examples/mcp_basic_agent/mcp_agent.secrets.yaml.example
+++ b/examples/mcp_basic_agent/mcp_agent.secrets.yaml.example
@@ -1,0 +1,7 @@
+$schema: ../../schema/mcp-agent.config.schema.json
+
+openai:
+  api_key: openai_api_key
+
+anthropic:
+  api_key: foo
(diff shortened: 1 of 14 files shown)


examples/mcp_basic_agent/mcp_agent.secrets.yaml.example
```diff
@@ -4,4 +4,4 @@ openai:
   api_key: openai_api_key
 
 anthropic:
-  api_key: anthropic_api_key
+  api_key: foo
```
saqadri: undo

saqadri (approved):
Thanks! One comment, otherwise looks good

examples/mcp_basic_agent/mcp_agent.secrets.yaml.example
evalstate: :) oops! fixed now.

saqadri:
Thank you @evalstate!! First contribution already :)

PR #4, opened 2025-01-22
Add Streamlit MCP RAG Agent example

StreetLamb:


examples/streamlit_mcp_rag_agent/main.py
```diff
@@ -0,0 +1,112 @@
...
+from mcp_agent.app import MCPApp
+from mcp_agent.agents.agent import Agent
+from mcp_agent.workflows.llm.augmented_llm_openai import OpenAIAugmentedLLM
+from agent_state import get_agent_state
+import streamlit as st
+
+
+def initialize_collection():
+    """Create and add data to collection."""
+    client = QdrantClient("http://localhost:6333")
```
saqadri, line 12: Does this URL matter? I think it should be fine to hardcode for example purposes, but maybe worth a comment to update

saqadri (approved):
Love it! Really nice work @StreetLamb

examples/streamlit_mcp_rag_agent/main.py
saqadri: Going to merge this in and we can update after if needed!

PR #7, opened 2025-01-25
Add MCP agent server example

StreetLamb:


examples/mcp_agent_server/server.py
```diff
@@ -0,0 +1,128 @@
+from mcp.server.models import InitializationOptions
```
saqadri, line 1: This is awesome. This might be something we put in the library itself

saqadri (approved):
Really great work! Approving so you can keep iterating on top of this. I'll merge in the changes to have filesystem based logger, and then update this example to use that

examples/mcp_agent_server/README.md
```diff
@@ -0,0 +1,27 @@
+# MCP Server example
+
+To use in claude desktop, insert the following in `claude_desktop_config.json` and restart the Claude Desktop client.
```
saqadri, line 3: Don't forget to add a screenshot or demo video embed in here!

src/mcp_agent/context.py
```diff
@@ -123,7 +127,7 @@ async def configure_logger(config: "Settings"):
     Configure logging and tracing based on the application config.
     """
     event_filter: EventFilter = EventFilter(min_level=config.logger.level)
-    print(f"Configuring logger with level: {config.logger.level}")
+    logger.info(f"Configuring logger with level: {config.logger.level}")
```
saqadri, line 130: Wow very interesting that this works. When I tried this before I got a circular dependency error. But I might have fixed it accidentally while refactoring other things

saqadri, line 130: Wow very interesting that this works. When I tried this before I got a circular dependency error. But I might have fixed it accidentally while refactoring other things

saqadri: If you notice any weird behavior feel free to remove this line entirely

saqadri:
This is a really really cool app and usecase @StreetLamb

PR #18, opened 2025-02-01
Progress Reporting Feature

evalstate:
Draft of Progress Capture. 

Intent was to build on existing Logging System and provide a convenient way to produce rich display of progress events. Design should be simple to be able to add/filter scopes, add/easily enrich UX/progress display mechanisms and later use "Progress" log event types for high volume ephemeral comms.

Main changes:

 - File Logger saves in JSONL format
 - Secrets are obscured by the JSON serializer unless environment variable overrides
 - Addition of ProgressLogger which captures the log event stream to convert log events to progress info
 - Addition of contextual data to important log events for progress reporting
 - Rudimentary formatter for progress events (on by default, can be switched off with config option `progress_display=False`)
 - Offline summarizer, viewer, rich progress stimulator for further development and testing of display.

Future enhancements could include
 - Use of "Progress" log events specifically to capture ephemeral high volume status updates (e.g. token streaming, MCP notifications). 
 - Abstractions in the augmented_llm areas etc. to avoid duplication of intent. Better use of constants/common methods to allow identification of important progress events.
 - Console redirection and better use of Rich to make it more visually appealing.
 - Better display of intent (e.g. Agent Call vs. Assistant Call) for Chatting.


saqadri:
So excited to see this! Just commenting that I will review this properly today

evalstate:
o3-mini support added.

evalstate:
mcp_researcher example added, support for passing environment vars to mcp servers

examples/mcp_basic_agent/main.py
```diff
@@ -53,10 +56,11 @@ async def example_usage():
             )
             logger.info(f"Result: {result}")
 
+        # Make sure to shut down logging cleanly
+        await LoggingConfig.shutdown()
```
saqadri: Should this be done in the async context manager that MCPApp sets up? And in `cleanup` in `app.py`

examples/mcp_hello_world/mcp_agent.config.yaml
```diff
@@ -1,17 +1,22 @@
...
     filesystem:
       command: "npx"
-      args:
-        [
-          "-y",
-          "@modelcontextprotocol/server-filesystem",
-          "/Users/saqadri/Desktop",
-          "/Users/saqadri/Downloads",
-        ]
+      args: ["-y", "@modelcontextprotocol/server-filesystem", "."]
```
saqadri, line 16: great idea to replace with '.'

examples/mcp_researcher/main.py
```diff
@@ -0,0 +1,67 @@
...
+
+app = MCPApp(name="mcp_root_test")
+
+
+async def example_usage():
+    async with app.run() as agent_app:
+        logger = agent_app.logger
+        context = agent_app.context
+
+        #        logger.info("Current config:", data=context.config.model_dump())
```
saqadri:
```suggestion
```

examples/mcp_model_selector/main.py
```diff
@@ -5,6 +5,7 @@
 from mcp_agent.app import MCPApp
 from mcp_agent.logging.logger import get_logger
 from mcp_agent.workflows.llm.llm_selector import ModelSelector
+from rich import print
```
saqadri, line 8: Is this needed here?

examples/mcp_basic_agent/main.py
```diff
@@ -1,10 +1,13 @@
 import asyncio
 import os
+import time
 
 from mcp_agent.app import MCPApp
 from mcp_agent.agents.agent import Agent
 from mcp_agent.workflows.llm.augmented_llm_anthropic import AnthropicAugmentedLLM
 from mcp_agent.workflows.llm.augmented_llm_openai import OpenAIAugmentedLLM
+from mcp_agent.logging.logger import LoggingConfig
+from rich import print
```
saqadri, line 9: I don't see this being used

examples/mcp_root_test/mcp_agent.config.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+mcp:
+  servers:
+    root_test:
+      command: "uv"
+      args: ["run", "root_test_server.py"]
+      roots:
+        - uri: "file:///home/ssmith/"
+          name: "test_data"
+          server_uri_alias: "file:///home/ssmith/"
```
saqadri: This is ok but just checking if it'll work for others

scripts/event_viewer.py
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env python3
```
saqadri, line 1: This is a very useful script, can't wait to try it out

src/mcp_agent/config.py
```diff
@@ -49,6 +72,12 @@ class MCPServerSettings(BaseModel):
     auth: MCPServerAuthSettings | None = None
     """The authentication configuration for the server."""
 
+    roots: Optional[List[MCPRootSettings]] = None
+    """Root directories this server has access to."""
+    
+    env: Dict[str, str] | None = None
```
saqadri, line 81: Thanks for adding this! I was going to add this so perhaps we can get your PR merged

src/mcp_agent/logging/json_serializer.py
```diff
@@ -15,16 +16,46 @@ class JSONSerializer:
...
+        'client_secret',
+        'access_token',
+        'refresh_token',
+    }
+
     def __init__(self):
         # Set of already processed objects to prevent infinite recursion
         self._processed_objects: Set[int] = set()
+        # Check if secrets should be logged in full
+        self._log_secrets = os.getenv('LOG_SECRETS', '').upper() == 'TRUE'
```
saqadri: Instead of checking LOG_SECRETS in env, this should probably be part of `LoggerSettings` so it can be specified in `mcp_agent.config.yaml`

src/mcp_agent/logging/listeners.py
```diff
@@ -85,11 +86,17 @@ async def handle_matched_event(self, event):
         }
         level = level_map.get(event.type, logging.INFO)
 
+        # Check if this is a server stderr message and format accordingly
+        if event.name == "mcpserver.stderr":
+            message = f"MCP Server: {event.message}"
```
saqadri, line 91: Does this have any more info on what server the message came from?

src/mcp_agent/logging/transport.py
```diff
@@ -67,36 +68,39 @@ class ConsoleTransport(FilteredEventTransport):
...
+        
         # Create namespace without None
         namespace = event.namespace
         if event.name:
             namespace = f"{namespace}.{event.name}"
 
         log_text = Text.assemble(
             (f"[{event.type.upper()}] ", style),
-            (f"{event.timestamp.isoformat()} ", "cyan"),
+            (f"{event.timestamp.replace(microsecond=0).isoformat()} ", "cyan"),
```
saqadri, line 96: Any reason for this?

src/mcp_agent/logging/transport.py
```diff
@@ -150,8 +154,8 @@ async def send_matched_event(self, event: Event) -> None:
 
         try:
             with open(self.filepath, mode=self.mode, encoding=self.encoding) as f:
-                # Write the log entry as JSON with newline
-                f.write(json.dumps(log_entry, indent=2) + "\n")
+                # Write the log entry as compact JSON (JSONL format)
```
saqadri, line 159: Good call

src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
```diff
@@ -127,6 +128,8 @@ async def generate(
             logger.debug(
                 f"Iteration {i}: Calling {model} with messages:",
                 data=messages,
+                model=model,
+                chat_turn=(len(messages) + 1) // 2,
```
saqadri: Should these be added to `data`? Or is `data` redundant now?

src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
```diff
@@ -197,7 +200,7 @@ async def generate(
         if params.use_history:
             self.history.set(messages)
 
-        logger.debug("Final response:", data=responses)
+        logger.debug(FINAL_RESPONSE_LOG_MESSAGE, data=responses, model=model)
```
saqadri: Same question as above

src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -136,19 +157,27 @@ async def generate(self, message, request_params: RequestParams | None = None):
...
+            if self._reasoning:
+                arguments = {
+                    **arguments,
+                    "max_completion_tokens": params.maxTokens,
+                    "reasoning_effort": self._reasoning_effort,
+                }
+            else:
+                arguments = {**arguments, "max_tokens": params.maxTokens}
+                if available_tools:
+                    arguments["parallel_tool_calls"] = params.parallel_tool_calls
```
saqadri, line 168: Should this be outside the if self._reasoning/else block?
```suggestion
            if available_tools:
                arguments["parallel_tool_calls"] = params.parallel_tool_calls
```

tests/test_event_progress.py
```diff
@@ -0,0 +1,70 @@
+"""Test event progress conversion from log events."""
```
saqadri, line 1: Tests for the feature! I love it

saqadri (approved):
@evalstate thank you so much for adding this! Lots of fantastic changes in this PR. I have a few questions about the specifics, but marking this as approved so we can merge this in when you are comfortable with the changes. If it doesn't regress any functionality then I would prefer to merge and build on top

examples/mcp_basic_agent/main.py
evalstate: removed and tested.

examples/mcp_model_selector/main.py
evalstate: yes - want rich console to handle interleaving.

examples/mcp_basic_agent/main.py
evalstate: removed loggingconfig, left rich.print for interleaving (not _strictly_ necessary but handy if somoeone wants to add print statements without breaking other rich features).

src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
evalstate: for later review. there is definitely a better way to handle the semantic info being passed around. in this case, IIRC i was thinking about keeping the keys separate to make it easier to distinguish what came from where.

src/mcp_agent/workflows/llm/augmented_llm_openai.py
evalstate: it should; o3 models 400 if you send this parameter.

tests/test_event_progress.py
evalstate: yes, there should be a few more. i've also added event_replay.py with some captured log files if anyone wants to create new/better rich_progress displays (which might have a window for last conversation message etc.).

saqadri:
@evalstate thank you for this incredible change. It's superb, and I will merge it in and we can fix-forward any minor issues from there!

PR #21, opened 2025-02-06
Support models compatible with OpenAI format

### LINKED ISSUE #19: ollama support
Could I use it with ollama, or using litellm that connects to ollama? If yes, wich model to choose? I can run up to 70b models locally? Or multi 32b models...

StreetLamb:
Expose openai's `base_url` in `mcp_agent.config.yaml` file so that providers using OpenAI format such as Ollama, LiteLLM, Gemini (in preview) and Deepseek etc. can be used without having to write custom classes for each of them. Included an example to demonstrate its usage. 

Closes #19

### DIFF
--- a/examples/mcp_basic_ollama_agent/README.md
+++ b/examples/mcp_basic_ollama_agent/README.md
@@ -1,0 +1,14 @@
+# MCP Ollama Agent example
+
+This example shows a "finder" Agent using llama models to access the 'fetch' and 'filesystem' MCP servers.
+
+You can ask it information about local files or URLs, and it will make the determination on what to use at what time to satisfy the request.
+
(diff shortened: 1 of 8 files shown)


saqadri (approved):
Amazing!!!

tihomir-kit:
Sorry to parachute in, but would this be the correct way to use this with the Gemini 2.5 Pro Preview?

`mcp_agent.config.yaml`

```
openai:
  base_url: "https://generativelanguage.googleapis.com/"
  default_model: "gemini-2.5-pro-preview-03-25"
  api_key: "<my_api_key>"
```

What did you use for base_url ?

Then I imagine you would simply use this:

```
llm = await finder_agent.attach_llm(OpenAIAugmentedLLM)
```

And it should just work?

PR #22, opened 2025-02-08
Fix duplicate tool call messages when using OpenAIAugmentedLLM 

StreetLamb:
When a tool call message has content, `OpenAIAugmentedLLM` will append the message twice. This fix ensures that messages are only added once. Also log and handle exceptions from model for both `OpenAIAugmentedLLM` and `AnthropicAugmentedLLM`.

### DIFF
--- a/src/mcp_agent/workflows/llm/augmented_llm_openai.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_openai.py
@@ -205,11 +205,6 @@ class OpenAIAugmentedLLM(
                             continue
                         if result is not None:
                             messages.append(result)
-            elif choice.finish_reason == "stop":
-                # We have reached the end of the conversation
-                messages.append(message)
-                logger.debug(f"Iteration {i}: Stopping because finish_reason is 'stop'")
-                break
             elif choice.finish_reason == "length":
                 # We have reached the max tokens limit
                 logger.debug(
@@ -223,6 +218,11 @@ class OpenAIAugmentedLLM(
                     f"Iteration {i}: Stopping because finish_reason is 'content_filter'"
                 )
                 # TODO: saqadri - would be useful to return the reason for stopping to the caller
+                break
+            else:  # choice.finish_reason == "stop"
+                # We have reached the end of the conversation
+                messages.append(message)
+                logger.debug(f"Iteration {i}: Stopping because finish_reason is 'stop'")
                 break
 
         if params.use_history:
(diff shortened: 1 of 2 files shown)


src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -163,65 +163,67 @@ async def generate(self, message, request_params: RequestParams | None = None):
...
-                    )
+                messages.append(converted_message)
 
                 if message.tool_calls:
-                    # Execute all tool calls in parallel
+                    # Execute all tool calls in parallel.
                     tool_tasks = [
                         self.execute_tool_call(tool_call)
-                        for tool_call in message.tool_calls
+                        for tool_call in choice.message.tool_calls
```
saqadri: why not `message`, since you assigned `message=choice.message` above?

src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -163,65 +163,67 @@ async def generate(self, message, request_params: RequestParams | None = None):
...
-                # We have reached the end of the conversation
-                logger.debug(f"Iteration {i}: Stopping because finish_reason is 'stop'")
-                break
-            elif choice.finish_reason == "length":
-                # We have reached the max tokens limit
-                logger.debug(
-                    f"Iteration {i}: Stopping because finish_reason is 'length'"
+            if choice.finish_reason in ["tool_calls", "function_call"]:
+                converted_message = self.convert_message_to_message_param(
+                    choice.message, name=self.name
```
saqadri:
```suggestion
                    message, name=self.name
```

src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -163,65 +163,67 @@ async def generate(self, message, request_params: RequestParams | None = None):
...
+                )
+                # TODO: saqadri - would be useful to return the reason for stopping to the caller
+                break
+            elif choice.finish_reason == "content_filter":
+                # The response was filtered by the content filter
+                logger.debug(
+                    f"Iteration {i}: Stopping because finish_reason is 'content_filter'"
+                )
+                # TODO: saqadri - would be useful to return the reason for stopping to the caller
+                break
```
saqadri: shouldn't we append `message` here as well?

Potentially consider having this line outside the if/elif block, and removing it from the tool_calls case so it's not added twice

```
converted_message = self.convert_message_to_message_param(
                    choice.message, name=self.name
)
messages.append(converted_message)
```

saqadri (approved):
Thank you!!

PR #24, opened 2025-02-11
WIP: Add scheduling assistant flagship project

StreetLamb:
Scheduling assistant MVP built with mcp-agent. The assistant will poll gmail inbox every minute and process incoming emails.

### DIFF
--- a/flagship/scheduling-assistant/main.py
+++ b/flagship/scheduling-assistant/main.py
@@ -1,0 +1,67 @@
+import asyncio
+from mcp_agent.app import MCPApp
+from mcp_agent.agents.agent import Agent
+from mcp_agent.workflows.llm.augmented_llm_openai import OpenAIAugmentedLLM
+from mcp_agent.workflows.llm.augmented_llm import RequestParams
+
+
+from prompts import SCHEDULING_AGENT_INSTRUCTIONS
+from servers.gmail import GmailService
+
+CRED_FILE_PATH = "/Users/jerronlim/Desktop/projects/mcp-agent-jerron/flagship/scheduling-assistant/creds/client_creds.json"
+TOKEN_PATH = "/Users/jerronlim/Desktop/projects/mcp-agent-jerron/flagship/scheduling-assistant/creds/gmail_app_tokens.json"
+
+app = MCPApp(name="mcp-scheduling-assistant")
+
+
+def generate_agent_prompt(thread_messages: list):
+    messages_str = str(thread_messages)
+    return f"""
+    Below is the full conversation thread of an email:
+    {messages_str}
+    
+    Please perform the necessary actions based on your instructions.
+    """
+
+
+async def main():
+    async with app.run() as mcp_agent_app:
+        logger = mcp_agent_app.logger
+
+        gmail_service = GmailService(CRED_FILE_PATH, TOKEN_PATH)
+
+        scheduling_agent = Agent(
+            name="scheduler",
+            instruction=SCHEDULING_AGENT_INSTRUCTIONS,
+            server_names=["calendar", "gmail"],
+        )
+
+        async with scheduling_agent:
+            llm = await scheduling_agent.attach_llm(OpenAIAugmentedLLM)
+
+            while True:
+                logger.info("Checking for unread emails.")
+                emails = await gmail_service.get_unread_emails()
+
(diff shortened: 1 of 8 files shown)


flagship/scheduling-assistant/main.py
```diff
@@ -0,0 +1,67 @@
...
+from mcp_agent.agents.agent import Agent
+from mcp_agent.workflows.llm.augmented_llm_openai import OpenAIAugmentedLLM
+from mcp_agent.workflows.llm.augmented_llm import RequestParams
+
+
+from prompts import SCHEDULING_AGENT_INSTRUCTIONS
+from servers.gmail import GmailService
+
+CRED_FILE_PATH = "/Users/jerronlim/Desktop/projects/mcp-agent-jerron/flagship/scheduling-assistant/creds/client_creds.json"
+TOKEN_PATH = "/Users/jerronlim/Desktop/projects/mcp-agent-jerron/flagship/scheduling-assistant/creds/gmail_app_tokens.json"
```
saqadri, line 12: nit: should we set this in env, or in the config.yaml itself? I believe we can add extras to any of the settings

flagship/scheduling-assistant/prompts.py
```diff
@@ -0,0 +1,45 @@
+SCHEDULING_AGENT_INSTRUCTIONS = """
```
saqadri, line 1: Whoa this is a very precise prompt. Does it work reliably/have you found cases where it doesn't?
Not important for now, but one thing we can consider is break this down into a series of steps and kind of strictly define the workflow that at any given time the LLM is expected to do.

For e.g. to extract meeting details, we could have a single llm.generate_structured call that returns meeting details in a pydantic object, which can then be used for further processing.

But if this works one-shot, then that's pretty great!

flagship/scheduling-assistant/servers/gmail.py
```diff
@@ -0,0 +1,421 @@
+#!/usr/bin/env python3
```
saqadri, line 1: Nice! You should consider contributing this server to https://github.com/modelcontextprotocol/servers, it's a solid reference implementation

saqadri (approved):
Thanks for getting this started! This is already in good enough shape to merge IMO and then we can keep building on top.

A few questions in addition to the PR comments:
1. So if I'm understanding correctly, the gmail creds are for the scheduling assistant's own google account, which is being polled for new emails.
2. Is the calendar creds for the user's google account? (i.e. we are giving the scheduling assistant direct access to the user's calendar?)

flagship/scheduling-assistant/servers/google_calendar.py
```diff
@@ -0,0 +1,221 @@
...
+            service = build("calendar", "v3", credentials=self.creds)
+            return service
+        except HttpError as error:
+            logger.error(f"Error initializing Calendar service: {error}")
+            raise ValueError(f"An error occurred: {error}")
+
+    async def list_events(self) -> dict:
+        try:
+            events_result = await asyncio.to_thread(
+                self.service.events().list, calendarId="primary"
```
saqadri, line 65: Should this accept a filter to be able to filter the events to a specific time period? For example, the agent might want to look at the calendar over the next week, or a particular day, to see availability.

flagship/scheduling-assistant/prompts.py
```diff
@@ -0,0 +1,45 @@
...
+     - Include all participants found in the email (To, CC, From), plus yourself.
+     - Add meeting agenda, location, and video link (if applicable).
+     - **Send a confirmation email** with meeting details.
+     - **Reply to the sender and keep those in CC in CC (excluding yourself if already in CC).**
+     - **Sign off as "Scheduling Assistant."**
+
+   - **If there is a conflict**:
+     - Identify and propose alternative slots based on availability and working hours.
+     - **Reply to the sender and keep those in CC in CC (excluding yourself if already in CC).**
+     - Upon confirmation, schedule the meeting and send invites.
```
saqadri, line 33: How would this "upon confirmation" part work, since the LLM would have to send the email first and end its turn. When it polls and finds a new email, it will concatenate the email thread into its context so it'll have memory of the state, but will it "resume" from this point onwards?

saqadri:
@StreetLamb closing this PR but keeping the branch around. This will be directly related to some of the application work kickstarting on top of mcp-agent shortly.

PR #26, opened 2025-02-13
Ensure base_url is always used for OpenAI clients

mcolyer:
It looks like there was at least one place where this wasn't happening, not sure if there are others.


saqadri (approved):
Ooof, great catch, thank you @mcolyer! I'll review if there's any others but I think you've got it

saqadri:
Will publish a patch today

saqadri:
@mcolyer I've published v0.0.7 which includes this fix. Thank you again for the contribution, and please let me know if there are any other requests or issues with mcp-agent that we should prioritize addressing

PR #28, opened 2025-02-20
fix finder_agent.py readme example

roterski:
Hello :wave:

I had just tried running the finder_agent example from the readme but it didn't work until I've fixed those two typos.

Thanks for the project!


saqadri:
Thanks for the great catch @roterski!

Issue #31, opened 2025-02-27
Settings for MCPApp programmatically


andersenthomas98:
Hi, really like this project! :)

I have a use-case for programmatically defining settings when using the `MCPApp` class. E.g.:

```python
from mcp_agent.app import MCPApp
from mcp_agent.agents.agent import Agent
from mcp_agent.config import Settings, MCPSettings, MCPServerSettings, OpenAISettings
import asyncio

async def main():
    config = Settings(
            mcp=MCPSettings(
                servers={
                    "fetch": MCPServerSettings(
                        command="uvx",
                        args=["mcp-server-fetch"]
                    )
                }
            )
        )

    app = MCPApp(name="mcp_basic_agent", settings=config)
    
    async with app.run():
        agent = Agent(
            name="mcp_basic_agent",
            instruction="Test",
            server_names=["fetch"]
        )
        async with agent:
            # Raises error here: ValueError: Server 'fetch' not found in registry  
            print("Agent connected to server")

if __name__ == "__main__":
    asyncio.run(main())
```
However, this raises an error when initializing the agent. This only works if there is a `mcp_agent.config.yaml` file with the following content: 
```yaml
mcp:
  servers:
    fetch:
      command: "uvx"
      args: ["mcp-server-fetch"]
```
I understand that this usage is not documented anywhere, however, according to the MCPApp interface this should be possible(?)

saqadri:
@andersenthomas98 thanks for reaching out! this looks like a bug, as it should definitely be possible to set the settings object programmatically on MCPApp. I'll take a look soon

saqadri:
@andersenthomas98 this fix has been published to pypi in v0.0.8. Thank you for contributing the fix!

dacox:
@andersenthomas98 @saqadri how do you specify `default_model` using `OpenAISettings` when setting `Settings` programatically? it does not appear to have the same structure as it does in the config yaml

saqadri:
> [@andersenthomas98](https://github.com/andersenthomas98) [@saqadri](https://github.com/saqadri) how do you specify `default_model` using `OpenAISettings` when setting `Settings` programatically? it does not appear to have the same structure as it does in the config yaml

@dacox it is done as follows:

https://github.com/lastmile-ai/mcp-agent/blob/main/examples/basic/mcp_basic_agent/main.py#L37

The `OpenAISettings` type accepts extra args, of which `default_model` can be one of them. I can add that as a named property so you get autocomplete when setting it up

PR #37, opened 2025-03-03
Store context globally on init

andersenthomas98:
Fixes [this](https://github.com/lastmile-ai/mcp-agent/issues/31) issue.


saqadri:
Thanks for the contribution @andersenthomas98! I'll merge this, test this and publish a patch. I am surprised this is the issue as I would have expected the config to still be set on the MCPApp instance programmatically without setting it globally

PR #39, opened 2025-03-05
Get Anthropic default model from config if available

MattMorgis:
I had this in my `mcp_agent.config.yaml` but my logs still indicated `claude-3-5` 

```yaml
# mcp_agent.config.yaml
anthropic:
  default_model: "claude-3-7-sonnet-latest"
```

### DIFF
--- a/src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
@@ -65,8 +65,14 @@ class AnthropicAugmentedLLM(AugmentedLLM[MessageParam, Message]):
             speedPriority=0.4,
             intelligencePriority=0.3,
         )
+
+        default_model = (
+            self.context.config.anthropic.default_model
+            if self.context.config.anthropic
+            else "claude-3-5-sonnet-20241022"
+        )
         self.default_request_params = self.default_request_params or RequestParams(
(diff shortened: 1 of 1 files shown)


src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
```diff
@@ -65,8 +65,14 @@ def __init__(self, *args, **kwargs):
             speedPriority=0.4,
             intelligencePriority=0.3,
         )
+
+        default_model = (
+            self.context.config.anthropic.default_model
+            if self.context.config.anthropic
+            else "claude-3-5-sonnet-20241022"
+        )
```
saqadri: Can you update it to this -- we can also now start defaulting to claude-3.7

```suggestion
        default_model = "claude-3-7-sonnet-latest"  # Fallback default

        if self.context.config.anthropic:
            if hasattr(self.context.config.anthropic, "default_model"):
                default_model = self.context.config.anthropic.default_model
```

saqadri (approved):
Thank you for the fix @MattMorgis! I literally noticed this yesterday and had a fix in a private branch, so thanks for fixing :). Just requested a change to make it more robust (in the case that `anthropic` settings are defined but not `default_model` is specified, then it should still default to claude-3.7

saqadri:
Amazing, thank you!!

PR #42, opened 2025-03-07
Multi transport logging 

MattMorgis:
Allows for a config like this:
```yaml
logger:
  transports: [file, console]
  level: debug
  path_pattern: "logs/mcp-agent-{timestamp}.jsonl"
  timestamp_format: "%Y%m%d_%H%M%S"
```

Keeps backward compatibility: 
```yaml
logger:
  type: file
  level: debug
  path: "logs/mcp-agent.jsonl"
```

### DIFF
--- a/src/mcp_agent/config.py
+++ b/src/mcp_agent/config.py
@@ -188,10 +188,12 @@ class LoggerSettings(BaseModel):
     """Path to log file, if logger 'type' is 'file'."""
 
     # Support for auto-generated filenames with timestamp patterns
-    path_pattern: str = "logs/mcp-agent-{timestamp}.jsonl"
+    path_pattern: str | None = None
     """
     Path pattern for file transport with timestamp placeholder.
     {timestamp} will be replaced with current time in format specified by timestamp_format.
+    If not specified, the standard 'path' setting will be used instead.
+    Example: "logs/mcp-agent-{timestamp}.jsonl"
     """
 
     timestamp_format: str = "%Y%m%d_%H%M%S"
--- a/src/mcp_agent/config.py
+++ b/src/mcp_agent/config.py
@@ -172,7 +172,11 @@ class LoggerSettings(BaseModel):
     Logger settings for the MCP Agent application.
     """
 
+    # Original transport configuration (kept for backward compatibility)
     type: Literal["none", "console", "file", "http"] = "console"
+
+    transports: List[Literal["none", "console", "file", "http"]] = ["console"]
+    """List of transports to use (can enable multiple simultaneously)"""
 
     level: Literal["debug", "info", "warning", "error"] = "info"
     """Minimum logging level"""
@@ -182,6 +186,19 @@ class LoggerSettings(BaseModel):
 
     path: str = "mcp-agent.jsonl"
     """Path to log file, if logger 'type' is 'file'."""
+
+    # Support for auto-generated filenames with timestamp patterns
+    path_pattern: str = "logs/mcp-agent-{timestamp}.jsonl"
+    """
+    Path pattern for file transport with timestamp placeholder.
+    {timestamp} will be replaced with current time in format specified by timestamp_format.
+    """
+
+    timestamp_format: str = "%Y%m%d_%H%M%S"
+    """Format string for timestamp in path_pattern"""
+
+    use_run_id: bool = False
+    """Whether to use a unique run ID in the filename instead of timestamp"""
 
     batch_size: int = 100
     """Number of events to accumulate before processing"""
--- a/src/mcp_agent/config.py
+++ b/src/mcp_agent/config.py
@@ -10,31 +10,6 @@ from pydantic import BaseModel, ConfigDict, field_validator, Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
-class LogPathSettings(BaseModel):
-    """
-    Settings for configuring log file paths with dynamic elements like timestamps or session IDs.
-    """
-    path_pattern: str = "logs/mcp-agent-{unique_id}.jsonl"
-    """
-    Path pattern for log files with a {unique_id} placeholder.
-    The placeholder will be replaced according to the unique_id setting.
-    Example: "logs/mcp-agent-{unique_id}.jsonl"
-    """
-    
-    unique_id: Literal["timestamp", "session_id"] = "timestamp"
-    """
-    Type of unique identifier to use in the log filename:
-    - timestamp: Uses the current time formatted according to timestamp_format
-    - session_id: Generates a UUID for the session
-    """
-    
-    timestamp_format: str = "%Y%m%d_%H%M%S"
-    """
-    Format string for timestamps when unique_id is set to "timestamp".
-    Uses Python's datetime.strftime format.
-    """
-
-
 class MCPServerAuthSettings(BaseModel):
     """Represents authentication configuration for a server."""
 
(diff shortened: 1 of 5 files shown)


src/mcp_agent/config.py
```diff
@@ -183,6 +187,21 @@ class LoggerSettings(BaseModel):
...
+    {timestamp} will be replaced with current time in format specified by timestamp_format.
+    If not specified, the standard 'path' setting will be used instead.
+    Example: "logs/mcp-agent-{timestamp}.jsonl"
+    """
+
+    timestamp_format: str = "%Y%m%d_%H%M%S"
+    """Format string for timestamp in path_pattern"""
+
+    use_run_id: bool = False
+    """Whether to use a unique run ID in the filename instead of timestamp"""
```
saqadri: I would suggest simplifying this by creating a `LogPathSettings` model, with the following fields:
```suggestion
    # Support for auto-generated filenames with timestamp patterns
    path_settings: LogPathSettings | None = None
    """
    Save log files with more advanced path semantics, like having timestamps or session id in the log name.
    """
```

```
class LogPathSettings(BaseModel):
    # TODO: add docstrings from before
    path_pattern: str = "logs/mcp-agent-{unique_id}.jsonl"
    unique_id: Literal["timestamp", "session_id"] = "timestamp"
    timestamp_format: str = "%Y%m%d_%H%M%S"
```

src/mcp_agent/logging/transport.py
```diff
@@ -432,19 +434,113 @@ async def _process_events(self):
...
+        Args:
+            event: Event to send
+        """
+        # Send to all transports, capturing exceptions but not failing.
+        exceptions = []
+        for transport in self.transports:
+            try:
+                await transport.send_event(event)
+            except Exception as e:
+                exceptions.append((transport, e))
```
saqadri: Might want to switch this to asyncio.gather(...) so you send to all the transports in parallel

src/mcp_agent/logging/transport.py
```diff
@@ -432,19 +434,113 @@ async def _process_events(self):
...
+    # If no pattern specified, use the static pattern
+    if (
+        not hasattr(settings, "path_pattern")
+        or "{timestamp}" not in settings.path_pattern
+    ):
+        return settings.path
+
+    # Generate timestamp or run ID
+    if hasattr(settings, "use_run_id") and settings.use_run_id:
+        replacement = str(uuid.uuid4())
```
saqadri: Can you create an issue for me to have a session_id added to the `Context` object on each MCPApp run... that will be helpful so we aren't generating this here.

src/mcp_agent/logging/transport.py
```diff
@@ -432,19 +434,113 @@ async def _process_events(self):
...
+                        batch_size=settings.batch_size,
+                        timeout=settings.http_timeout,
+                        event_filter=event_filter,
+                    )
+                )
+        if transports:
+            return MultiTransport(transports)
+        else:
+            return NoOpTransport(event_filter=event_filter)
+    # Backward compatibility for single transport configuration
```
saqadri: A couple of issues I noticed:

1. if someone specifies just one item in `transports` then we should use that single transport the same way if someone specified `transport`.
2. 'path_pattern' only seems to apply for multiple transports, but if someone specifies just transport type it won't ever apply.
3. Looks like some raise ValueError were changed to warnings. I actually think it's better to raise if some transports are misconfigured.

Also, there's a bit of code duplication happening so we can simplify the entire function as follows:

    transports: List[EventTransport] = []

    transport_types = []

    if hasattr(settings, "transports"):
        transport_types = settings.transports
    else:
        transport_types = [settings.type]
    
    for transport_type in transport_types:
        if transport_type == "none":
                continue
        elif transport_type == "console":
            transports.append(ConsoleTransport(event_filter=event_filter))
        elif transport_type == "file":
            filepath = get_log_filename(settings) # IMPORTANT: needs to include
            #if settings.path:
            #   return settings.path
            #elif settings.path_settings:
            #    <current logic for generate_log_filename>
            #else:
            #    raise ValueError("File path required for file transport. Either specify 'path' or 'path_settings'")
            
            transports.append(
                FileTransport(filepath=filepath, event_filter=event_filter)
            )
        elif transport_type == "http":
            if not settings.http_endpoint:
                raise ValueError("HTTP endpoint required if HTTP transport is specified")
            transports.append(
                HTTPTransport(
                    endpoint=settings.http_endpoint,
                    headers=settings.http_headers,
                    batch_size=settings.batch_size,
                    timeout=settings.http_timeout,
                    event_filter=event_filter,
                )
            )
        else:
            raise ValueError(f"Unsupported transport type: {settings.type}")#
        
    if not transports:
        return NoOpTransport(event_filter=event_filter)
    elif len(transports) > 1:
        return MultiTransport(transports)
    else:
        return transports[0]

saqadri (changes requested):
Thanks for adding this @MattMorgis! Directionally looks good to me, but I have a few suggestions which would be great to incorporate before merging. Please let me know if you have any questions on them, and thanks again!

src/mcp_agent/logging/transport.py
```diff
@@ -432,19 +434,113 @@ async def _process_events(self):
...
+                exceptions.append((transport, e))
+
+        # Log any exceptions that occurred
+        if exceptions:
+            print(f"Errors occurred in {len(exceptions)} transports:")
+            for transport, exc in exceptions:
+                print(f"  {transport.__class__.__name__}: {exc}")
+
+
+def generate_log_filename(settings: LoggerSettings) -> str:
```
saqadri: Some comments in other parts of the change would affect this function, so just flagging for completeness:

This function should take into account if `path` is specified, in which case it should just return that

```
#if settings.path:
        #   return settings.path
        #elif settings.path_settings:
        #    <current logic for generate_log_filename>
        #else:
        #    raise ValueError("File path required for file transport. Either specify 'path' or 'path_settings'")
 
```

src/mcp_agent/logging/transport.py
mattmorgis: #47

MattMorgis:
@saqadri I think I addressed all of your feedback. I also updated the README and the basic agent example to reflect the new config options. Let me know if you need anything else!

src/mcp_agent/config.py
```diff
@@ -183,6 +213,12 @@ class LoggerSettings(BaseModel):
     path: str = "mcp-agent.jsonl"
     """Path to log file, if logger 'type' is 'file'."""
 
+    # Settings for advanced log path configuration
+    path_settings: Optional["LogPathSettings"] = None
```
saqadri, line 217: Out of curiosity, any reason for referencing the type as `"LogPathSettings"` instead of `LogPathSettings`? Could do `LogPathSettings | None` if you drop the quotations. Just curious, not a big deal

src/mcp_agent/logging/transport.py
```diff
@@ -432,30 +434,123 @@ async def _process_events(self):
...
+            unique_id = str(uuid.uuid4())
+        else:  # timestamp is the default
+            now = datetime.datetime.now()
+            time_format = settings.path_settings.timestamp_format
+            unique_id = now.strftime(time_format)
+
+        return path_pattern.replace("{unique_id}", unique_id)
+
+    # Default fallback
+    return settings.path
```
saqadri, line 504: Perhaps at this point we should raise ValueError because we should have been able to generate something valid by this point?

src/mcp_agent/config.py
mattmorgis: Ha, nice catch. A combination of Claude doing it + me still new enough at Python to not notice. I can submit a follow up, the rest of the file uses the `| None` pattern

Issue #51, opened 2025-03-11
Downgrading the minimum numpy version to 2.1.3.


Nicba1010:
Is it possible to downgrade the minimum numpy version to 2.1.3? I am having issues integrating mcp-agent into our project because of our dependency on numba, which hasn't yet upgraded to numpy 2.2.

I have no idea if this is even a valid request to make.

saqadri:
> Is it possible to downgrade the minimum numpy version to 2.1.3? I am having issues integrating mcp-agent into our project because of our dependency on numba, which hasn't yet upgraded to numpy 2.2.
> 
> I have no idea if this is even a valid request to make.

@Nicba1010 let me try, I think that should be ok!

saqadri:
Thanks for submitting the fix! I'll push a patch update today to pypi

Nicba1010:
Thank you for merging it that quickly!

PR #52, opened 2025-03-11
Relax numpy dependency constraing, now requiring numpy>=2.1.3

Nicba1010:
As per #51 

Unfortunately we cannot use mcp-agent because of our numba dependency. This would resolve it. And it doesn't seem to affect anything.


saqadri (approved):
Thanks @Nicba1010! I just used latest numpy but I'm using pretty basic functionality from it so downgrading is ok

PR #59, opened 2025-03-13
fix example streamlit_mcp_rag_agent with last mcp-server-qdrant 

venticello:
> From [v0.7.0](https://github.com/qdrant/mcp-server-qdrant/releases/tag/v0.7.0) Command parameters are no longer supported. All the configuration has to be done with environmental variables.

Qdrant MCP server parameters are moved from arg to env.


saqadri (approved):
Thank you for the fix, @venticello!

Issue #47, opened 2025-03-09
Generate Session IDs


MattMorgis:
Currently in the logging module, unique IDs (UUIDs) are generated for log files when `use_run_id` is enabled. It would be better to have a consistent session ID that's set at the MCPApp level and shared across the application context.

This would provide a few benefits:
- Consistent ID for tracing a single run across logs and other systems
- Reusable identifier for correlating events from the same session
- Avoid duplicating UUID generation in multiple places

The session_id should be:
- Generated once at the beginning of an MCPApp run
- Added to the Context object
- Available throughout the application
- Used for log filenames and other identifiers (update logging to use this ID when available)

PR #63, opened 2025-03-16
feat: add session_id

corroleaus:
fix: #47 

- Added session_id generation in app.py
- propagate to context and logger
- use for log filenames
- possible to extend usage as needed
- checks to ensure session_id exists on context and logger
- ran format and lint


src/mcp_agent/logging/logger.py
```diff
@@ -255,21 +265,26 @@ async def managed(cls, **config_kwargs):
...
     """
 
     with _logger_lock:
+        # Create a new logger if one doesn't exist
         if namespace not in _loggers:
-            _loggers[namespace] = Logger(namespace)
+            _loggers[namespace] = Logger(namespace, session_id)
+        # If we have a session_id and the existing logger doesn't, update it
+        elif session_id is not None and _loggers[namespace].session_id is None:
+            _loggers[namespace].session_id = session_id
```
saqadri: What scenario is this necessary under?

src/mcp_agent/context.py
```diff
@@ -183,9 +189,12 @@ async def initialize_context(
     context.config = config
     context.server_registry = ServerRegistry(config=config)
 
+    # Use the provided session ID or generate a new one
+    context.session_id = session_id or str(uuid.uuid4())
```
saqadri: Should we just leave it as session_id here to make it deterministic? i.e. if someone doesn't specify one we don't generate one.

saqadri (approved):
@corroleaus this is fantastic. Thank you for this. I reviewed and it looks good for the most part. The only concern I have is we are autogen'ing the session id in multiple places, which may lead to unexpected behavior. Left some comments regarding that. Other that that, looks good!

src/mcp_agent/logging/logger.py
corroleaus: Yea i don't see the need for this either, I'll fix that

src/mcp_agent/context.py
corroleaus: Yea, i think we can just do `context.session_id = session_id` instead and remove the uuid import

corroleaus: Fixed

src/mcp_agent/logging/logger.py
corroleaus: Fixed

saqadri:
Great work @corroleaus! This has been merged and I'll publish a patch update tomorrow

PR #70, opened 2025-03-24
readme: fix swarm agent example

sayanarijit:


saqadri (approved):
Thankk you for the great catch!

PR #71, opened 2025-03-24
Fix structure and code block issue in README

skyduy:
1. The structure of contents and `Table of Contents` does not match.
2. There is an issue with the python code block in `Persistent server connections`.


saqadri (approved):
Good catch! Thank you @skyduy

PR #95, opened 2025-04-01
fix: update tool result content formatting in AnthropicAugmentedLLM

### LINKED ISSUE #94: Image Tool Use Result Causes Response Error in Anthropic API
I was trying to return an image to Anthropic as a tool use result, but the message sent to the Anthropic API looks like this, which results in a response error:
```javascript
{
'role': 'user', 
'content': [{
        'type': 'tool_result', 
        'tool_use_id': 'toolu_0189y67pid1MQxkr6ytTH21h',
        'content': [ImageContent(type='image', data='iVBORw0K...........==', mimeType='image/png', annotations=None)], 
        'is_error': False,
  }]
}
```

I found that in augmented_llm_anthropic.py, after calling tools, the type of result.content added to the messages list does not seem to match the expected format:

```python
result = await self.call_tool( 
    request=tool_call_request, tool_call_id=tool_use_id
)

messages.append(
    MessageParam(
        role="user",
        content=[
            ToolResultBlockParam(
                type="tool_result",
                tool_use_id=tool_use_id,
                content=result.content,   
                // result.content is list[TextContent | ImageContent | EmbeddedResource]
                // Expected : Union[str, Iterable[Content]], where Content is Union[TextBlockParam, ImageBlockParam]
                is_error=result.isError,
            )
        ],
    )
)
```

Is there a missing type conversion for result.content before adding it to the message? Or Could AnthropicMCPTypeConverter already be handling this conversion elsewhere without me knowing?

Thanks a lot!

StreetLamb:
Fixes #94. Previously MCP server response from tool calling was not converted to Anthropic's ToolResultBlockParam content type.

### DIFF
--- a/src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
@@ -41,6 +41,7 @@ from mcp_agent.workflows.llm.augmented_llm import (
     MCPMessageResult,
     ProviderToMCPConverter,
     RequestParams,
+    CallToolResult,
 )
 from mcp_agent.logging.logger import get_logger
 
@@ -232,29 +233,9 @@ class AnthropicAugmentedLLM(AugmentedLLM[MessageParam, Message]):
                             request=tool_call_request, tool_call_id=tool_use_id
                         )
 
-                        tool_result_content: list[
-                            Union[TextBlockParam, ImageBlockParam]
-                        ] = []
-                        for content_block in result.content:
-                            converted_content = mcp_content_to_anthropic_content(
-                                content_block, for_message_param=True
-                            )
-                            if converted_content["type"] in ["text", "image"]:
-                                tool_result_content.append(converted_content)
+                        message = self.from_mcp_tool_result(result, tool_use_id)
 
-                        messages.append(
-                            MessageParam(
-                                role="user",
-                                content=[
-                                    ToolResultBlockParam(
-                                        type="tool_result",
-                                        tool_use_id=tool_use_id,
-                                        content=tool_result_content,
-                                        is_error=result.isError,
-                                    )
-                                ],
-                            )
-                        )
+                        messages.append(message)
 
         if params.use_history:
             self.history.set(messages)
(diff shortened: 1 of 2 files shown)


src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
```diff
@@ -211,14 +212,48 @@ async def generate(
...
+                                    return TextBlockParam(
+                                        type="text", text=content_block.resource.text
+                                    )
+                                else:
+                                    # Best effort to convert since this is not supported by anthropic ToolResultBlockParam.
+                                    return TextBlockParam(
+                                        type="text",
+                                        text=f"{content_block.resource.mimeType}:{content_block.resource.blob}",
+                                    )
+
```
saqadri: Should this logic be in or utilize the `AnthropicMCPTypeConverter` class below?

saqadri (approved):
Great fix, @StreetLamb! Wondering about the proper place for this, but otherwise lgtm

src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
StreetLamb: I was planning to leverage on `from_mcp_message_param` method, but it expects a MessageParam containing a single content block, whereas the `call_tool` returns a result with a list of content blocks so they are not really compatible.

saqadri: What if we loop through the list of content blocks returned by the `call_tool` and use the `from_mcp_message_param` method?

Alternatively, we could add a `from_mcp_tool_call` method to the type converter. I feel like this would be a common enough issue across all model providers. But lmk what you think -- happy to merge in as is as well

StreetLamb: Added a new `from_mcp_tool_result` method to handle tool results from MCP servers. Let me know what you think!

saqadri (approved):
Awesome, thanks! Do we need to update any other AugmentedLLM classes as well for this?

PR #96, opened 2025-04-01
Create server.py

JowaFlux:


saqadri:
@JowaFlux looks like you might have created the PR too early. If you're trying to expose the MCPApp and its agents as an MCP Server, rest assured that I'm working on that as we speak!

PR #102, opened 2025-04-02
Add support for AWS profile in Bedrock settings

### LINKED ISSUE #101: AWS Profile for Bedrock Augumented LLM
There can be multiple AWS profiles and credentials stored in ~/.aws/credentials
There should be provision to select which aws profile for which we need to read credentials and connect to aws bedrock

StreetLamb:
Resolves #101

### DIFF
--- a/src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
@@ -1,5 +1,5 @@
 from typing import List, Type
-from boto3 import client
+from boto3 import Session
 from mcp.types import (
     CallToolRequestParams,
     CallToolRequest,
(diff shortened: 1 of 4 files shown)


src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
```diff
@@ -55,15 +55,17 @@ def __init__(self, *args, **kwargs):
                 default_model = self.context.config.bedrock.default_model
 
         if self.context.config.bedrock:
-            self.bedrock_client = client(
+            session = Session(profile_name=self.context.config.bedrock.profile)
```
saqadri, line 58: Does this work ok when profile is unspecified ?

saqadri (approved):
Nice, thanks for the quick fix @StreetLamb!!

src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
StreetLamb: Yup it will use the default profile if unspecified

Issue #25, opened 2025-02-13
content in OpenAI ChatCompletion messages is a list, not string


xiaolei0125:
use home assistant as mcp server,  then Calling OpenAI ChatCompletion with messages,  but the content in message is a list structure.

OpenAI API Specification：
- content can only be a string or null.
- It cannot be a list, object, or any other type.

If content is a list,  some LLM(deepseek) return the following error:
`[mcp_agent.workflows.llm.augmented_llm_openai] Error: Failed to deserialize the JSON body into the target type: messages[2]: invalid type: sequence, expected a string at line 1 column 639`

```
[DEBUG] 2025-02-12T17:25:30.871390 mcp_agent.workflows.llm.augmented_llm_openai - Iteration 0: Tool call results: [{'role': 'tool', 'tool_call_id': 
'call_e9ZCGCWE7PmQIap9IwOOZkZV', 'content': [{'type': 'text', 'text': '{"speech": {}, "response_type": "action_done", "data": {"targets": [], "success": 
[{"name": "DC3\\u5f00\\u51732", "type": "entity", "id": "switch.dc1_d3_sw2"}], "failed": []}}'}]}]
[DEBUG] 2025-02-12T17:25:30.871501 mcp_agent.workflows.llm.augmented_llm_openai - Iteration 1: Calling OpenAI ChatCompletion with messages:
{
  "data": [
    {
      "role": "system",
      "content": "for home assistant tools, do not use device_class, use name directly"
    },
    {
      "role": "user",
      "content": "关闭充电开关"
    },
    {
      "role": "assistant",
      "content": null,
      "tool_calls": [
        {
          "id": "call_e9ZCGCWE7PmQIap9IwOOZkZV",
          "function": {
            "arguments": "{\"name\":\"充电开关\"}",
            "name": "homeassistant-HassTurnOff"
          },
          "type": "function"
        }
      ],
      "audio": null,
      "refusal": null,
      "name": "finder"
    },
    {
      "role": "tool",
      "tool_call_id": "call_e9ZCGCWE7PmQIap9IwOOZkZV",
      "content": [
        {
          "type": "text",
          "text": "{\"speech\": {}, \"response_type\": \"action_done\", \"data\": {\"targets\": [], \"success\": [{\"name\": \"DC3\\u5f00\\u51732\", 
\"type\": \"entity\", \"id\": \"switch.dc1_d3_sw2\"}], \"failed\": []}}"
        }
      ]
    }
  ]
}
[...]

PR #106, opened 2025-04-03
Fix OpenAI ChatCompletion message content format issue

askme765cs:
- Modified `execute_tool_call` method in OpenAIAugmentedLLM to convert content list to string
- Used "\n".join() to concatenate list items into a single string
- This addresses issue #25 where OpenAI API rejects list-type content
- Follows OpenAI API specification requiring content to be string or null
- Prevents errors with LLMs like deepseek that strictly validate input format

The change ensures compatibility with OpenAI's API requirements which specify: "content can only be a string or null. It cannot be a list, object, or any other type."


saqadri (approved):
Thanks for your contribution, @askme765cs! Merging this in, though there may be a wider fix that may be needed to follow the pattern from @StreetLamb's latest fix in the Anthropic class.

PR #116, opened 2025-04-06
feature: support websocket connections and passing in HTTP headers to websockets and SSE connections

pmohan6:
Add support for connecting over websockets
Add support for passing in HTTP headers to websocket and SSE connections.
Tested against a real websocket server hosted on smithery.ai. Added an example.
Tested passing headers against a personally hosted SSE server. Added an example.

Added unit tests for websockets.

You can run unit tests as follows:
```
uv run pytest tests/test_websocket.py
```


saqadri (approved):
Apologies for the delay in getting to this @pmohan6. This is an excellent PR, and can be merged as-is! Thank you for driving this!

PR #117, opened 2025-04-07
fix: refactor assistant message arguments in OpenAIAugmentedLLM

softage0:
This change ensures that the code follows the API reference.

- refactor assistant role arguments of chat completion messages in OpenAIAugmentedLLM to follow [OpenAI API reference](https://platform.openai.com/docs/api-reference/chat/create)
  - `content` and `tool_calls` are optional but not null
- fix code format by lint.py


saqadri (approved):
Thank you @softage0, appreciate your attention to detail!

PR #119, opened 2025-04-07
Update mcp_basic_azure_agent README.md

lucinvitae:
Solution for `ModuleNotFoundError` when setting up this example.


saqadri (approved):
Thank you @lucinvitae, good catch!

PR #120, opened 2025-04-08
Fix: Functions Passed to Agents Are Now Available as Tools

rholinshead:
## Summary
Currently, functions passed to agents do not properly get listed as tools since the async `initialize` is not (always?) called. To fix, pull the `_function_tool_map` construction into `__init__`.

## Changes
- Construct `_function_tool_map` in `__init__` in `agent.py`
- Add `functions` example to show working usage

## Test Plan
```
uv run scripts/example.py run functions --debug
```

See our functions listed as tools:
```
[INFO] 2025-04-07T20:15:13 mcp_agent.mcp_agent_using_functions - Tools available:
{
  "data": {
    "meta": null,
    "nextCursor": null,
    "tools": [
      {
        "name": "add_numbers",
        "description": "\n    Adds two numbers.\n    ",
        "inputSchema": {
          "properties": {
            "a": {
              "title": "A",
              "type": "integer"
            },
            "b": {
              "title": "B",
              "type": "integer"
            }
          },
          "required": [
            "a",
            "b"
          ],
          "title": "add_numbersArguments",
          "type": "object"
        }
      },
      {
        "name": "multiply_numbers",
        "description": "\n    Multiplies two numbers.\n    ",
        "inputSchema": {
          "properties": {
            "a": {
              "title": "A",
              "type": "integer"
            },
            "b": {
              "title": "B",
              "type": "integer"
            }
          },
          "required": [
            "a",
            "b"
          ],
          "title": "multiply_numbersArguments",
          "type": "object"
        }
      },
```

See them being called as tools:
```
...
 {
        "finish_reason": "tool_calls",
        "index": 0,
        "logprobs": null,
        "message": {
          "content": null,
          "refusal": null,
          "role": "assistant",
          "annotations": [],
          "audio": null,
          "function_call": null,
          "tool_calls": [
            {
              "id": "call_fp1JQviBV6p4Yfbb2gU6pASf",
              "function": {
                "arguments": "{\"a\":2,\"b\":3}",
                "name": "add_numbers"
              },
              "type": "function"
            }
          ]
        }
      }
...

 {
        "finish_reason": "tool_calls",
        "index": 0,
        "logprobs": null,
        "message": {
          "content": null,
          "refusal": null,
          "role": "assistant",
          "annotations": [],
          "audio": null,
          "function_call": null,
          "tool_calls": [
            {
              "id": "call_P4LwnJKUtYQ8wjKLxMo5JIt7",
              "function": {
                "arguments": "{\"a\":5,\"b\":4}",
                "name": "multiply_numbers"
              },
              "type": "function"
            }
          ]
        }
      }
...
```

And correct response:
```
[INFO] 2025-04-07T20:15:16 mcp_agent.mcp_agent_using_functions - Expert math result: The result of adding 2 and 3 is 5. 
When you multiply that result by 4, you get 20.
```

### DIFF
--- a/examples/functions/main.py
+++ b/examples/functions/main.py
@@ -1,0 +1,62 @@
+import asyncio
+import time
+
+from mcp_agent.app import MCPApp
+from mcp_agent.agents.agent import Agent
+from mcp_agent.workflows.llm.augmented_llm_openai import OpenAIAugmentedLLM
+
+
+def add_numbers(a: int, b: int) -> int:
+    """
+    Adds two numbers.
+    """
+    print(f"Math expert is adding {a} and {b}")
+    return a + b
+
+
+def multiply_numbers(a: int, b: int) -> int:
+    """
+    Multiplies two numbers.
+    """
+    print(f"Math expert is multiplying {a} and {b}")
+    return a * b
+
+
+app = MCPApp(name="mcp_agent_using_functions")
+
+
+async def example_usage():
+    async with app.run() as agent_app:
+        logger = agent_app.logger
+        context = agent_app.context
+        logger.info("Current config:", data=context.config.model_dump())
+
+        math_agent = Agent(
+            name="math_agent",
+            instruction="""You are an expert in mathematics with access to some functions
+            to perform correct calculations. 
+            Your job is to identify the closest match to a user's request, 
+            make the appropriate function calls, and return the result.""",
+            functions=[add_numbers, multiply_numbers],
+        )
+
+        async with math_agent:
+            logger.info("math_agent: Connected to server, calling list_tools...")
+            result = await math_agent.list_tools()
+            logger.info("Tools available:", data=result.model_dump())
+
+            llm = await math_agent.attach_llm(OpenAIAugmentedLLM)
+            result = await llm.generate_str(
+                message="Add 2 and 3, then multiply the result by 4.",
+            )
+
+            logger.info(f"Expert math result: {result}")
+
+
+if __name__ == "__main__":
+    start = time.time()
+    asyncio.run(example_usage())
+    end = time.time()
+    t = end - start
+
+    print(f"Total run time: {t:.2f}s")
--- a/src/mcp_agent/agents/agent.py
+++ b/src/mcp_agent/agents/agent.py
@@ -64,6 +64,9 @@ class Agent(MCPAggregator):
 
         # Map function names to tools
         self._function_tool_map: Dict[str, FastTool] = {}
+        for function in self.functions:
+            tool: FastTool = FastTool.from_function(function)
+            self._function_tool_map[tool.name] = tool
 
         self.human_input_callback: HumanInputCallback | None = human_input_callback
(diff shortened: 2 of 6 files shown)


examples/functions/main.py
```diff
@@ -0,0 +1,62 @@
...
+        math_agent = Agent(
+            name="math_agent",
+            instruction="""You are an expert in mathematics with access to some functions
+            to perform correct calculations. 
+            Your job is to identify the closest match to a user's request, 
+            make the appropriate function calls, and return the result.""",
+            functions=[add_numbers, multiply_numbers],
+        )
+
+        async with math_agent:
```
saqadri, line 43: This should initialize the agent

src/mcp_agent/agents/agent.py
```diff
@@ -79,10 +82,6 @@ async def initialize(self):
             self.__aenter__()
```
saqadri: It would be good to change the initialize function to get rid of this, and instead to implement the aenter asynccontextmanager for this class, which a) calls super.aenter, and b) calls initialize.

I think for now initialize can stay in empty and just call super.initialize()

src/mcp_agent/agents/agent.py
```diff
@@ -64,6 +64,9 @@ def __init__(
 
         # Map function names to tools
         self._function_tool_map: Dict[str, FastTool] = {}
+        for function in self.functions:
```
saqadri, line 67: This is a good change to have in the constructor.

saqadri (approved):
Thanks for the fix @rholinshead! I think your fix uncovered another issue which is that initialize shouldn't be calling aenter, but rather an aenter should be calling initialize. That mistake is why this wasn't working as intended.

src/mcp_agent/agents/agent.py
rholinshead: Ok, I looked into this and:
- confirmed that implementing `__aenter__` and calling initialize with the function tool map construction in initialize (to test) worked as expected, e.g.
```
    async def __aenter__(self):
        """
        Initialize the agent and connect to the MCP servers.
        NOTE: This method is called automatically when the agent is used as an async context manager.
        """
        if self.initialized:
            return self

        await super().__aenter__()
        await self.initialize()
        return self

    async def initialize(self):
        for function in self.functions:
            tool: FastTool = FastTool.from_function(function)
            self._function_tool_map[tool.name] = tool
```

But, having the function tool map construction in `__init__` IMO makes sense (sounds like we agree). Looks like there is no `initialize` in any superclass so `super().initialize()` is invalid. In that case, we'll end up with an empty `initialize` and then `__aenter__` becomes a redundant override.
So, might make sense to just remove `__aenter__` and `initialize` from this class for now and add back in the future if needed?

rholinshead: @saqadri lastest version here removes them as redundant. If you'd prefer to have them just to be explicit / to allow future updates to initialize please let me know and I can add them back

saqadri (approved):
thanks!

PR #131, opened 2025-04-13
Fix generate_structured output in AzureAugmentedLLM and from_mcp_message_param issues

StreetLamb:


src/mcp_agent/workflows/llm/augmented_llm_azure.py
```diff
@@ -266,9 +266,10 @@ async def generate_structured(
         )
         request_params.metadata = metadata
 
-        structured_response = await self.generate_str(
-            message=message, request_params=request_params
-        )
+        response = await self.generate(message=message, request_params=request_params)
+        json_data = json.loads(response[-1].content)
```
saqadri, line 270: Is this guaranteed to be valid json?

saqadri (approved):

saqadri (approved):

PR #133, opened 2025-04-14
Feature/llm response history

### LINKED ISSUE #132: surfacing `response_history` attribute from `AugmentedLLMs`
Hello, 

I think there is an opportunity in surfacing the `response_history` attribute of `AugmentedLLMs`. 

This would open up other analytics such as token consumption, comparison between providers, performance monitoring, model behavior patterns, and other performance metrics.

We have a PR in the works with @felixbouazza on the subject in that would be of interest :)

felixbouazza:
Hello,

This pull request introduces a new feature for the MCP agent by exposing the `response_history` attribute in the `AugmentedLLMs` class. This update is a foundational improvement that supports enhanced observability into LLM operations. By making this attribute accessible, we enable a range of valuable analytics that can drive better insights and performance monitoring. 

close #132

### DIFF
--- a/examples/mcp_agent_token_usage_tracking/README.md
+++ b/examples/mcp_agent_token_usage_tracking/README.md
@@ -1,0 +1,10 @@
+# Tracking Token Usage with AugmentedLLMs
+
+This example demonstrates how to monitor token consumption across different LLM providers using the new `response_history` attribute of AugmentedLLMs.
+
+## What you'll learn
+- How to access and analyze token usage data
+- Compare token efficiency between OpenAI and Anthropic models for identical queries
+- Leverage the `response_history` property for advanced analytics
+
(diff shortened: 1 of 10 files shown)


saqadri:
Oh this is quite interesting, @felixbouazza, thanks for this change. I am open to accepting this contribution as-is. Just give me a couple more days to think about this in relation to some other changes that are currently in-flight, such as updating AugmentedLLMs and the workflows to support Temporal execution.

The other thing that I need to think about is whether AugmentedLLM class should use the model provider's types (as they do today), or if they should natively use MCP types.

felixbouazza:
Hi @saqadri,
Thank you very much for your positive feedback and for taking the time to review this PR! 
We completely understand needing a few days to reflect on these changes.

Issue #138, opened 2025-04-14
Error with model selection with openai reasoning models


Pythonista7:
I notice that `OpenAIAugmentedLLM` on init sets the `self._reasoning` flag , the default fallback model is `gpt-4o` but when i specifically pass a model_name which is a **reasoning** model during `generate_str(request_params=...)` it fails with the below error.

```
[mcp_agent.workflows.llm.augmented_llm_openai.writing_assistant] Error: Error code: 400 - {'error': {'message': "Unsupported parameter: 
'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens',
 'code': 'unsupported_parameter'}}
```

I believe we would need to update the `self._reasoning` after the `self.select_model()` returns (line: 154) here? 

![Image](https://github.com/user-attachments/assets/0b1bc821-0cbd-4d7c-be44-a73ab066a027)

If there is an alternate way to resolve this or if a code change would be required would be happy to implement them based on suggestion :) Thanks!

StreetLamb:
Hi @Pythonista7, thanks for raising the issue! If you would like to, another thing to resolve would be changing `max_tokens` to `max_completion_tokens` since [max_tokens has been deprecated](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens).

Pythonista7:
@StreetLamb great suggestion! Have update the PR with the change: https://github.com/lastmile-ai/mcp-agent/pull/139

PR #139, opened 2025-04-15
fx: _reasoning flag used to set api params

Pythonista7:
# Changes
- re-evaluate `_reasoning` flag after model selection inside `generate()`

# Issue
This resolves https://github.com/lastmile-ai/mcp-agent/issues/138


saqadri (approved):
Thanks for the fix @Pythonista7!

PR #147, opened 2025-04-20
feat: Add deepseek support

jingx8885:
feat: add deepseek support.
fix bugs:
1.  what mcp_content_to_openai_content return is dict, so use dict read operation
2. raise exception when execute_tool_call failed to trigger  re-execute.
3. close openai_client when generate finish.


saqadri:
@jingx8885 thank you for contributing this! Is this for using deepseek with the deepseek API? Can you also please contribute an example to the repo (under `examples`) showing this being used?

jingx8885:
ok, I wil add some examples

saqadri:
@jingx8885 thank you! You can just update the existing PR if you prefer

PR #153, opened 2025-04-24
feat: add GLM series models supports and agent example

mcpflow:
…mplementation


mcpflow:
Supports the GLM text series models to call MCP, including the free model glm-flashx-250414.

PR #159, opened 2025-04-26
fix: Refactor AsyncEventBus: Defer loop and _queue initialization unt…

jingx8885:
…il start() to avoid binding the event loop on module import. Added checks to prevent emitting events or stopping the bus if it has not been started.


src/mcp_agent/mcp/stdio.py
```diff
@@ -133,23 +134,84 @@ async def stdin_writer():
...
-                )
+                logger.error(f"Error during process cleanup for PID {pid}: {e}")
+            finally:
+                # Ensure anyio resources are properly cleaned up
+                # This might be redundant if process.wait() succeeded, but good practice
+                try:
+                    await process.aclose()
+                except Exception as close_exc:
+                    logger.warning(
+                        f"Error closing anyio process resources for PID {pid}: {close_exc}"
```
saqadri: This is fairly complex -- can you please explain the reason for this change? 
Also, has this been tested on Windows?

saqadri (changes requested):
Thanks for the contribution! Mocving the loop creation to `start` instead of `__init__` makes a lot of sense. I don't understand the need for the changes to stdio.py, however. Unless there's a good reason for those changes we can remove those.

src/mcp_agent/mcp/stdio.py
jingx8885: When running the MCP in stdio mode on Windows using the command uv run mcp run mcp_server.py, the uv process acts as the parent process, launching the mcp.exe process, which in turn starts a Python process.

Originally, using process.terminate() would only kill the uv process, leaving the mcp.exe and Python processes as zombie processes, which is highly risky. To fix this, I switched to using psutil to perform cascading termination of the parent and child processes, effectively resolving the issue. This solution has been successfully tested on Windows.

```mcp config yaml
mcp:
  servers:
    mcp_server:
      command: uv
      args:
        - run
        - mcp
        - run
        - mcp_server.py
```

jingx8885: If the event loop is created in __init__, it will be triggered during the get_logger function is called. For example in mcp_server_registry line 30-34:
```python 
from mcp_agent.logging.logger import get_logger
from mcp_agent.mcp.mcp_connection_manager import MCPConnectionManager
from mcp_agent.mcp.websocket import websocket_client

logger = get_logger(__name__)
```
This means that as soon as most modules in the MCP package are imported, the loop will be created immediately, making its initialization uncontrollable.

By moving the loop creation to the start method, we ensure that the loop is tied to the actual process startup behavior rather than being triggered upon module import. This provides better control over when the loop is initialized.

jingx8885:
#163 this feature is to fix issue163

src/mcp_agent/app.py
```diff
@@ -192,11 +183,25 @@ async def run(self):
                 # App is initialized here
                 pass
         """
+        original_policy = None  # 用于存储原始策略
+        policy_changed = False  # 标记策略是否被修改
+        # Set event loop policy for Windows if needed
+        if sys.platform == "win32":
+            original_policy = asyncio.get_event_loop_policy()
+            # 仅当当前策略不是 Proactor 时才设置
```
saqadri: Can you please update the comments to be English (consistent with rest of codebase)

jingx8885: already updated in English

PR #161, opened 2025-04-26
feat: Update OpenAI client to AsyncOpenAI and add tool call support

jingx8885:
- Changed OpenAI client from synchronous version to asynchronous `AsyncOpenAI`.
- Added `need_tool_calls` parameter to `RequestParams` to enable tool call inclusion in responses.
- Optimized code formatting for consistency and readability.

### DIFF
--- a/src/mcp_agent/workflows/llm/augmented_llm.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm.py
@@ -129,6 +129,10 @@ class RequestParams(CreateMessageRequestParams):
     """
     The likelihood of the model selecting higher-probability options while generating a response.
     """
+    need_tool_calls: bool = True
+    """
+    Whether to include tool calls in the response.
+    """
 
 
 class AugmentedLLMProtocol(Protocol, Generic[MessageParamT, MessageT]):
--- a/src/mcp_agent/workflows/llm/augmented_llm_openai.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_openai.py
@@ -2,7 +2,7 @@ import json
 import re
 from typing import Iterable, List, Type
 
-from openai import OpenAI
+from openai import AsyncOpenAI
 from openai.types.chat import (
     ChatCompletionAssistantMessageParam,
     ChatCompletionContentPartParam,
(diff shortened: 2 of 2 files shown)


src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -110,7 +110,7 @@ async def generate(self, message, request_params: RequestParams | None = None):
         Override this method to use a different LLM.
         """
         config = self.context.config
-        openai_client = OpenAI(
+        openai_client = AsyncOpenAI(
```
saqadri, line 113: For my understanding what is the benefit of using AsyncOpenAI over OpenAI?

src/mcp_agent/workflows/llm/augmented_llm.py
```diff
@@ -129,6 +129,10 @@ class RequestParams(CreateMessageRequestParams):
     """
     The likelihood of the model selecting higher-probability options while generating a response.
     """
+    need_tool_calls: bool = True
```
saqadri, line 132: What's the use-case where we don't want to provide tools to the model? RequestParams is intended to be as close to the MCP spec as possible, so I would prefer to avoid adding additional properties unless there's a strong reason

saqadri (changes requested):
Thanks for the contribution! Let's remove the `need_tool_calls` property

src/mcp_agent/workflows/llm/augmented_llm_openai.py
jingx8885: DeprecationWarning: pytest-asyncio detected an unclosed event loop when tearing down the event_loop
  fixture: <ProactorEventLoop running=False closed=False debug=False>
  pytest-asyncio will close the event loop for you, but future versions of the
  library will no longer do so. In order to ensure compatibility with future
  versions, please make sure that:
      1. Any custom "event_loop" fixture properly closes the loop after yielding it
      2. The scopes of your custom "event_loop" fixtures do not overlap
      3. Your code does not modify the event loop in async fixtures or tests

    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

In my test case, if use OpenAI, unclosed event loop problem wil happen when run AsyncioExecutor._execute_task, soI try to change to AsyncOpenAI to solve this problem.

src/mcp_agent/workflows/llm/augmented_llm.py
jingx8885: In my usage, sometimes letting the LLM decide whether to call the MCP can be error-prone. So in cases where it's obvious, I can determine through keywords that the LLM clearly doesn't need to call the MCP. In such situations, I turn this switch on, which significantly improves the stability of collaboration between the LLM and MCP.

PR #173, opened 2025-05-01
Add ability to configure the http_client for openai

blevz:


saqadri (approved):
Thanks for adding this!

PR #181, opened 2025-05-09
Fixing absolute path of server in basic/decorator/yaml

Deathn0t:
When going to `cd examples/basic/decorator` and running the main script with `python main.py` it failed because of a wrong path to the launch server. This PR fixes it.


saqadri (approved):
Thank you for fixing! This was due to a recent refactoring of the examples directory

PR #187, opened 2025-05-13
fix: Update _get_next_step method in orchestrator.py

jingx8885:
- Fix message parameter name typo
- Ensure proper request_params handling in next step generation
- Maintain consistency with other method signatures


saqadri (approved):
Good fix @jingx8885 thank you!

PR #188, opened 2025-05-13
Fix parallel workflow schema in mcp_agent_server basic example

danlester:
The schema is incorrect - the required property needed to be promoted to a slightly higher level.


saqadri (approved):
Thank you!

saqadri:
@danlester appreciate the fix! Heads up I will be merging an update to the library this week which will include a much simpler way for exposing MCP agents as MCP servers

PR #190, opened 2025-05-13
add support for Azure Identity instead of keys

demoray:


src/mcp_agent/workflows/llm/augmented_llm_azure.py
```diff
@@ -84,13 +85,23 @@ def __init__(self, *args, **kwargs):
...
+                    credential=AzureKeyCredential(self.context.config.azure.api_key),
+                    **self.context.config.azure.model_dump(
+                        exclude={"endpoint", "credential"}
+                    ),
+                )
+            else:
+                self.azure_client = ChatCompletionsClient(
+                    endpoint=self.context.config.azure.endpoint,
+                    credential=DefaultAzureCredential(),
+                    credential_scopes=["https://cognitiveservices.azure.com/.default"],
```
saqadri: Should these credential scopes be configurable in the Azure settings config?

demoray: I've pushed an update that makes that suggestion.

saqadri (approved):
Thank you @demoray!

PR #192, opened 2025-05-13
fix: Conditionally include toolConfig in BedrockAugmentedLLM arguments

### LINKED ISSUE #185: When MCPAggregator.list_tools() returns empty tool list, calling bedrock Converse API will raise 'invalid length' ParamValidator error
### mcp_agent.workflows.llm.augmented_llm_bedrock -> line:108-159

if no tools returned from mcp server, the `toolConfig` for Bedrock Converse API will have a empty list which will not pass botocore's parameter validation, and a validation error raises:

```
Error: Parameter validation failed: Invalid length for parameter toolConfig.tools, value: 0, valid min length: 1
```

here are the original code:
```
response = await self.aggregator.list_tools()

tool_config: ToolConfigurationTypeDef = {
    "tools": [
        {
            "toolSpec": {
                "name": tool.name,
                "description": tool.description,
                "inputSchema": {"json": tool.inputSchema},
            }
        }
        for tool in response.tools
    ],
    "toolChoice": {"auto": {}},
}

responses: list[MessageUnionTypeDef] = []
model = await self.select_model(params)

for i in range(params.max_iterations):
    inference_config = {
        "maxTokens": params.maxTokens,
        "temperature": params.temperature,
        "stopSequences": params.stopSequences or [],
    }

    system_content = [
        {
            "text": self.instruction or params.systemPrompt,
        }
    ]

    arguments: ConverseRequestTypeDef = {
        "modelId": model,
        "messages": messages,
        "system": system_content,
        "inferenceConfig": inference_config,
        "toolConfig": tool_config,
    }

    if params.metadata:
        arguments = {
            **arguments,
            "additionalModelRequestFields": params.metadata,
        }

    self.logger.debug(f"{arguments}")
    self._log_chat_progress(chat_turn=(len(messages) + 1) // 2, model=model)

    executor_result = await self.executor.execute(
        self.bedrock_client.converse, **arguments
    )
```

StreetLamb:
Fixes #185

### DIFF
--- a/src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
@@ -141,8 +141,10 @@ class BedrockAugmentedLLM(AugmentedLLM[MessageUnionTypeDef, MessageUnionTypeDef]
                 "messages": messages,
                 "system": system_content,
                 "inferenceConfig": inference_config,
-                "toolConfig": tool_config,
             }
+
+            if len(tool_config["tools"]) > 0:
+                arguments["toolConfig"] = tool_config
 
             if params.metadata:
                 arguments = {
--- a/src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
@@ -143,7 +143,7 @@ class BedrockAugmentedLLM(AugmentedLLM[MessageUnionTypeDef, MessageUnionTypeDef]
                 "inferenceConfig": inference_config,
             }
 
-            if len(tool_config["tools"]) > 0:
+            if isinstance(tool_config["tools"], list) and len(tool_config["tools"]) > 0:
                 arguments["toolConfig"] = tool_config
 
             if params.metadata:


src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
```diff
@@ -141,9 +141,11 @@ async def generate(self, message, request_params: RequestParams | None = None):
                 "messages": messages,
                 "system": system_content,
                 "inferenceConfig": inference_config,
-                "toolConfig": tool_config,
             }
 
+            if len(tool_config["tools"]) > 0:
```
saqadri: Will len() work if the tool_config["tools"] is None?

saqadri (approved):

src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
saqadri: Would be good to just make this a bit more robust otherwise

saqadri (approved):

src/mcp_agent/workflows/llm/augmented_llm_bedrock.py
StreetLamb: Based on its type, tool_config["tools"] cannot be None since its of type `Sequence[ToolTypeDef]`. However, i included a more robust check just in case.

saqadri (approved):

PR #194, opened 2025-05-14
fix get prompt no return

zmetry:


saqadri (approved):

saqadri:
Awesome! Thank you for the contribution @zmetry

PR #198, opened 2025-05-15
Fix blocking issue of other asynchronous tasks when running generate_…

ohahahahua:
…structured

Resolved an issue where the execution of the `generate_structured` function was causing other asynchronous tasks to be blocked. This fix ensures that `generate_structured` operates without hindering the performance and execution of concurrent asynchronous operations.

Specifically, the following changes were made:
- Introduced non-blocking I/O operations for `generate_structured` to prevent it from occupying the event loop.
- Optimized internal task scheduling to ensure fair distribution of resources among all running tasks.
- Added unit tests to verify that `generate_structured` does not block other asynchronous tasks under various load conditions.

These adjustments will help improve the overall responsiveness and efficiency of our application when handling multiple concurrent tasks.

### DIFF
--- a/src/mcp_agent/workflows/llm/augmented_llm_openai.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_openai.py
@@ -337,7 +337,7 @@ class OpenAIAugmentedLLM(
 
         try:
             # Extract structured data from natural language
-            structured_response = async_client.chat.completions.create(
+            structured_response = await async_client.chat.completions.create(
                 model=model or "gpt-4o",
                 response_model=response_model,
                 messages=[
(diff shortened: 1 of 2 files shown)


src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -357,7 +357,7 @@ async def generate_structured(
                 mode=instructor.Mode.JSON,
             )
 
-            structured_response = client.chat.completions.create(
+            structured_response = async_client.chat.completions.create(
```
saqadri: don't we need to await this now?

src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -337,7 +337,7 @@ async def generate_structured(
 
         try:
             # Extract structured data from natural language
-            structured_response = client.chat.completions.create(
+            structured_response = async_client.chat.completions.create(
```
saqadri: Same here, we should await this now

saqadri (changes requested):
Thanks @ohahahahua, just make sure the generation is being awaited now. Also please describe how you tested the changes.

saqadri (approved):

PR #227, opened 2025-05-20
feat: Implement memory sharing for EvaluatorOptimizerLLM

jingx8885:
feat: Implement memory sharing for EvaluatorOptimizerLLM
- Add a shared memory parameter to the EvaluatorOptimizerLLM class.
- Implement the `share_memory_from` method to enable memory sharing functionality.

This change aims to optimize memory management between the evaluator and optimizer.

For example, the `optimizer_llm` is often bound to an MCP server. After invoking the MCP and receiving content, the LLM might sometimes "hallucinate" or produce responses inconsistent with the MCP's returned content. The `evaluator_llm`, when performing its assessment, needs access to the `optimizer_llm`'s memory to better determine if it is hallucinating or generating unexpected output. MeanWhile, optimizer_llm can do better work when having evaluator_llm's memory.

## Summary by CodeRabbit

- **New Features**
  - Added an option to enable shared memory between the evaluator and optimizer, allowing them to reference the same history when desired.

### DIFF
--- a/src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
+++ b/src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
@@ -150,7 +150,7 @@ class EvaluatorOptimizerLLM(AugmentedLLM[MessageParamT, MessageT]):
         else:
             raise ValueError(f"Unsupported evaluator type: {type(evaluator)}")
         if share_memory:
-            self.evaluator_llm.share_memory_from(self.optimizer_llm)
+            self.evaluator_llm.history = self.optimizer_llm.history
         self.min_rating = min_rating
         self.max_refinements = max_refinements
 
--- a/src/mcp_agent/workflows/llm/augmented_llm.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm.py
@@ -274,26 +274,6 @@ class AugmentedLLM(ContextDependent, AugmentedLLMProtocol[MessageParamT, Message
         self.model_selector = self.context.model_selector
         self.type_converter = type_converter
 
-    def share_memory_from(self, other: "AugmentedLLM"):
-        """
-        Share memory from another AugmentedLLM instance.
-
-        This creates a reference to the other instance's history, meaning both
-        instances will share the same memory object and any changes to history
-        in either instance will be reflected in both.
-
-        Args:
-        other: The AugmentedLLM instance to share memory from
-
-        Raises:
-        ValueError: If other is None or not an AugmentedLLM instance
-        """
-        if other is None:
-            raise ValueError("Cannot share memory from None")
-        if not isinstance(other, AugmentedLLM):
-            raise ValueError("Can only share memory from another AugmentedLLM instance")
-        self.history = other.history
-
     @abstractmethod
     async def generate(
         self,


jingx8885:
@saqadri

src/mcp_agent/workflows/llm/augmented_llm.py
```diff
@@ -274,6 +274,26 @@ def __init__(
         self.model_selector = self.context.model_selector
         self.type_converter = type_converter
 
+    def share_memory_from(self, other: "AugmentedLLM"):
```
saqadri: I would prefer not to add this to the main API, since someone can just do

```
llm.history = other_llm.history
```

themselves

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
```diff
@@ -147,7 +149,8 @@ def __init__(
             )
         else:
             raise ValueError(f"Unsupported evaluator type: {type(evaluator)}")
-
+        if share_memory:
+            self.evaluator_llm.share_memory_from(self.optimizer_llm)
```
saqadri: This is very interesting @jingx8885, but I'm curious if this works well in practice. 

I understand your point that the evaluator should consider the full interaction to be able to determine the quality. But does it catch the issues you mentioned in the PR description when it has access to the full history? Also, does the generator get confused in follow-up interactions, or does it work well?

saqadri (changes requested):
Sorry for the delay @jingx8885, and thanks for the contribution. In general supportive of this change, but I have questions of whether this works well in practice. Also would appreciate adding a test for this scenario as part of your change.

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
jingx8885: In my project, since the optimizer_llm calls the MCP to retrieve data and complete tasks, it is prone to hallucinations during task execution—generating results inconsistent with MCP data. Therefore, I need the evaluator_llm to verify whether the optimizer_llm adheres to the MCP data, which requires the evaluator_llm to access the optimizer_llm's historical information. I consider this scenario highly common, so I have implemented a shared memory functionality.

jingx8885: I strongly recommend adding this feature, as it is essential for effectively enhancing the capabilities of the evaluator_optimizer. @saqadri

saqadri: @jingx8885 can you please resolve my PR feedback to get this ready for merging. Specifically, remove the updates to AugmentedLLM class (see other comment)

jingx8885: sorry, got it!

amr01060 (approved):

src/mcp_agent/workflows/llm/augmented_llm.py
amr01060: > I would prefer not to add this to the main API, since someone can just do
> 
> ```
> llm.history = other_llm.history
> ```
> 
> themselves

amr01060: > > I would prefer not to add this to the main API, since someone can just do
> > 
> > ```
> > llm.history = other_llm.history
> > ```
> > 
> > themselves
> 
>

amr01060: > > I would prefer not to add this to the main API, since someone can just do
> > 
> > ```
> > llm.history = other_llm.history
> > ```
> > 
> > themselves
> 
>

amr01060: > > I would prefer not to add this to the main API, since someone can just do
> > 
> > ```
> > llm.history = other_llm.history
> > ```
> > 
> > themselves
> 
>

amr01060: > > I would prefer not to add this to the main API, since someone can just do
> > 
> > ```
> > llm.history = other_llm.history
> > ```
> > 
> > themselves
> 
>

PR #233, opened 2025-05-22
Distributed Tracing for Key Features: Agents, (asyncio) Workflows, AugmentedLLMs

rholinshead:
## Summary
Implement tracing for some key features of the project -- Wgents, asyncio Workflows and AugmentedLLMs. This configures a `TracingConfig` similar to the `LoggingConfig`, which sets up the global tracer provider with exporters based on the config/settings. `console`, `file` and `otlp` are supported. For now, the tracer is propagated within context but we should ultimately consider pulling it out to an app-associated global singleton tracer to decouple it from context to allow it to be more easily reused outside of the project internals.

We should consider this more of a starting point to get tracing working and unblock trace-dependent projects. Some next steps are:
- decouple from context
- generalize different span types for recording consistent data
- handle trace context propagation across MCP server request/response boundary
- add tracing to any remaining internals

## Testing
- Added a couple examples to check exhaustive tracing of AugmentedLLM and Agent methods,
```
uv run scripts/example.py run tracing/llm
uv run scripts/example.py run tracing/agent
```

- For asyncio workflows, tested each applicable workflow with tracing enabled and with it disabled to ensure correct behaviour
```
uv run scripts/example.py run basic/mcp_basic_agent
uv run scripts/example.py run workflows/workflow_evaluator_optimizer
uv run scripts/example.py run workflows/workflow_intent_classifier
uv run scripts/example.py run workflows/workflow_orchestrator_worker
uv run scripts/example.py run workflows/workflow_parallel
uv run scripts/example.py run workflows/workflow_router
```

## Summary by CodeRabbit

- **New Features**
  - Comprehensive OpenTelemetry tracing added across agents, LLMs (OpenAI, Anthropic, Azure), orchestrators, routers, parallel/fan-in/fan-out workflows, embeddings, and intent classifiers.
  - Enhanced tracing configuration with support for multiple exporters (console, file, OTLP), dynamic trace file paths, and advanced exporter settings.
  - New example tracing-enabled agent and LLM workflows with configuration, secrets templates, and documentation.
  - Introduced semantic conventions for GenAI and MCP telemetry attributes.
  - Added a file-based span exporter for trace output.
  - New tracing configuration class centralizing setup and enabling session-aware tracing.

- **Improvements**
  - Detailed span annotation capturing request, response, internal states, and error reporting for improved observability.
  - Centralized and session-aware tracing configuration with enriched telemetry across core components and workflows.
  - Enhanced tracing in tool calls, generation loops, classification, and orchestration for better debugging insights.
  - Added tracing spans to executor methods and client session communications.
  - Improved error handling and span status reporting in LLM workflows.
  - Added tracing to model selection and agent lifecycle methods.

- **Documentation**
  - Added and updated READMEs and example configs with instructions for enabling tracing and integrating with Jaeger.

- **Style/Refactor**
  - Formatting improvements and import cleanup for clarity and maintainability.
  - Localized imports to reduce overhead and dependencies.

- **Chores**
  - Updated dependencies and requirements to support new tracing and telemetry capabilities.


examples/tracing/agent/README.md
saqadri, line 1: Awesome examples!

examples/tracing/llm/requirements.txt
```diff
@@ -0,0 +1,7 @@
+# Core framework dependency
+mcp-agent @ file://../../../  # Link to the local mcp-agent project root
+
+# Additional dependencies specific to this example
```
saqadri, line 4: Might be worth importing the packages directly, unless this works. Worried it might install the `mcp-agent` package published on pypi

examples/workflows/workflow_evaluator_optimizer/mcp_agent.config.yaml
```diff
@@ -23,3 +23,11 @@ mcp:
...
   # Secrets (API keys, etc.) are stored in an mcp_agent.secrets.yaml file which can be gitignored
   default_model: gpt-4o
+
+otel:
+  enabled: false
+  exporters: ["console"]
+  # If running jaeger locally, uncomment the following lines and add "otlp" to the exporters list
+  # otlp_settings:
+  #   endpoint: "http://localhost:4318/v1/traces"
+  service_name: "WorkflowEvaluatorOptimizerExample"
```
saqadri, line 33: Is this needed? Or is this so that someone can just toggle enabled to `true` and get tracing for the example? If so, would be great to add a comment either here on in readme for each example.

examples/workflows/workflow_intent_classifier/mcp_agent.config.yaml
```diff
@@ -18,3 +18,11 @@ mcp:
 openai:
   # Secrets (API keys, etc.) are stored in an mcp_agent.secrets.yaml file which can be gitignored
   default_model: "gpt-4o-mini"
+
+otel:
+  enabled: false
+  exporters: ["console"]
```
saqadri, line 24: Would this work if the logger is also configured to console export? Do both of them export to console?

src/mcp_agent/app.py
```diff
@@ -236,10 +237,13 @@ async def run(self):
                 pass
         """
         await self.initialize()
-        try:
-            yield self
-        finally:
-            await self.cleanup()
+
+        tracer = self.context.tracer or trace.get_tracer("mcp-agent")
```
saqadri: `trace.get_tracer("mcp-agent")`

Should "mcp-agent" be `self.name` by chance?

src/mcp_agent/config.py
```diff
@@ -219,26 +219,65 @@ class UsageTelemetrySettings(BaseModel):
...
-    """Log spans to console"""
-
     sample_rate: float = 1.0
     """Sample rate for tracing (1.0 = sample everything)"""
 
+    otlp_settings: TraceOTLPSettings | None = None
+    """OTLP settings for OpenTelemetry tracing. Required if using otlp exporter."""
+
+    # Settings for advanced trace path configuration for file exporter
+    path_settings: TracePathSettings | None = None
```
saqadri, line 276: nit: should path_settings be in `TraceOTLPSettings`? Not a big deal either way

src/mcp_agent/executor/workflow.py
saqadri, line 1: Should this be traced as well? Perhaps in subsequent PR?

src/mcp_agent/mcp/mcp_agent_client_session.py
```diff
@@ -131,31 +144,80 @@ async def send_request(
...
+                result = await super().send_request(
+                    request,
+                    result_type,
+                    request_read_timeout_seconds,
+                    metadata,
+                    progress_callback,
+                )
+                logger.debug("send_request: response=", data=result.model_dump())
+
+                record_attributes(span, result.model_dump(), "result")
```
saqadri: nit: both of these are doing a `model_dump()` we should consolidate into a var

src/mcp_agent/mcp/mcp_agent_client_session.py
```diff
@@ -131,31 +144,80 @@ async def send_request(
...
-                request_read_timeout_seconds,
-                metadata,
-                progress_callback,
-            )
-            logger.debug("send_request: response=", data=result.model_dump())
-            return result
-        except Exception as e:
-            logger.error(f"send_request failed: {e}")
-            raise
+        tracer = self.context.tracer or trace.get_tracer("mcp-agent")
```
saqadri: should we have a self.context.get_tracer() function, or generally a global get_tracer(context: "Context") function which does `return context.tracer or trace.get_tracer("mcp-agent")`

Seems like this line is duplicated everywhere

src/mcp_agent/mcp/mcp_agent_client_session.py
```diff
@@ -131,31 +144,80 @@ async def send_request(
...
+                span.set_attribute(MCP_REQUEST_ID, str(related_request_id))
+            if notification.root.params:
+                record_attributes(
+                    span, notification.root.parms.model_dump(), MCP_REQUEST_ARGUMENT_KEY
+                )
+
+            try:
+                return await super().send_notification(notification, related_request_id)
+            except Exception as e:
+                logger.error("send_notification failed", data=e)
```
saqadri, line 227: Does the exception automatically get logged in the tracer?

src/mcp_agent/mcp/mcp_aggregator.py
```diff
@@ -122,103 +126,129 @@ def __init__(
...
+                        await connection_manager.__aenter__()
+                        self.context._mcp_connection_manager = connection_manager
 
-                self._persistent_connection_manager = connection_manager
+                    self._persistent_connection_manager = connection_manager
 
-        await self.load_servers()
-        self.initialized = True
+            await self.load_servers()
+            span.add_event("initialize_complete")
```
saqadri:
```suggestion
            span.add_event("initialized")
```

examples/workflows/workflow_evaluator_optimizer/mcp_agent.config.yaml
rholinshead: It's not needed, just for easy toggling to see the workflow traces. I can update the readmes in each case

examples/workflows/workflow_intent_classifier/mcp_agent.config.yaml
rholinshead: Yes, they can both export to console without issue

src/mcp_agent/app.py
rholinshead: The intention of using "mcp-agent" is mainly so that there's a consistent tracer name for all callers if for some reason tracer is not set in context; in practice I think that only happens when tracing is disabled, in which case the global tracer object is a ProxyTracer which does no-op spans. So I don't know if it really matters in practice what the fallback name is, but would still prefer to keep it a single consistent tracer name so that only one tracer would be used

src/mcp_agent/config.py
rholinshead: path_settings is specifically for the filepath whereas otlp settings are for external otlp collectors (like jaeger)

src/mcp_agent/mcp/mcp_aggregator.py
```diff
@@ -374,183 +451,331 @@ async def get_server(self, server_name: str) -> Optional[ClientSession]:
...
-            )
+            span.set_attribute("prompts", [prompt.name for prompt in res.prompts])
 
-        return ListPromptsResult(
-            prompts=[
-                namespaced_prompt.prompt.model_copy(
-                    update={"name": namespaced_prompt_name}
+            for prompt in res.prompts:
+                span.set_attribute(
+                    f"prompt.{prompt.name}.description", prompt.description
```
saqadri: worried about cases where prompt.description isn't available.

src/mcp_agent/executor/workflow.py
rholinshead: Ya, I think we can make a followup issue for it. I was focusing on asyncio only and not touching temporal workflows

src/mcp_agent/mcp/mcp_aggregator.py
```diff
@@ -374,183 +451,331 @@ async def get_server(self, server_name: str) -> Optional[ClientSession]:
...
+                    for attr in [
+                        "description",
+                        "required",
+                    ]:
+                        value = getattr(arg, attr, None)
+                        if value is not None:
+                            span.set_attribute(
+                                f"prompt.{prompt.name}.arguments.{arg.name}.{attr}",
+                                value,
+                            )
```
saqadri: Again worried if prompt.arguments is null. As well as any further sub-properties.. Ok to leave, but would harden the checks more

src/mcp_agent/tracing/tracer.py
```diff
@@ -0,0 +1,159 @@
...
+            OpenAIInstrumentor().instrument()
+        except ModuleNotFoundError:
+            logger.error(
+                "OpenAI otel instrumentation not available. Please install opentelemetry-instrumentation-anthropic."
+            )
+
+        cls._initialized = True
+
+
+# TODO: Figure out how to implement this with agnostic context manager for sync/async
```
saqadri: Needed?

src/mcp_agent/workflows/embedding/embedding_openai.py
```diff
@@ -25,21 +32,32 @@ def __init__(
...
+            span.set_attribute("data", data)
+            span.set_attribute("embedding_dim", self.embedding_dim)
+
+            response = self.client.embeddings.create(
+                model=self.model, input=data, encoding_format="float"
+            )
+
+            span.set_attribute(GEN_AI_RESPONSE_MODEL, response.model)
+            span.set_attribute(GEN_AI_USAGE_INPUT_TOKENS, response.usage.prompt_tokens)
+            span.set_attribute("gen_ai.usage.total_tokens", response.usage.total_tokens)
```
saqadri: Need checks similar to in the cohere example:

if response.meta and response.meta.tokens:
                if response.meta.tokens.input_tokens:
                    span.set_attribute(
                        GEN_AI_USAGE_INPUT_TOKENS, response.meta.tokens.input_tokens
                    )
                if response.meta.tokens.output_tokens:
                    span.set_attribute(
                        GEN_AI_USAGE_OUTPUT_TOKENS, response.meta.tokens.output_tokens
                    )

src/mcp_agent/mcp/mcp_agent_client_session.py
rholinshead: Ya, I can pull into a function. I'll do the global within telemetry.py since long-term I think it can be decoupled from context

src/mcp_agent/mcp/mcp_agent_client_session.py
rholinshead: Yes, any uncaught exception raised within a span will automatically get tracked as an exception event and set the span status to error

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
```diff
@@ -158,116 +162,189 @@ async def generate(
...
-            )
+            best_response = response
+            if isinstance(response, list) and len(response) > 0:
+                record_attributes(
+                    span,
+                    self.optimizer_llm.extract_response_message_attributes_for_tracing(
+                        response[0]
+                    ),
+                    "initial_response",
+                )
```
saqadri: I don't quite understand the response[0]

examples/tracing/llm/requirements.txt
rholinshead: Will update to directly. Fwiw, this does work as expected but I'm not sure I can guarantee it doesn't use the public version

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
```diff
@@ -158,116 +162,189 @@ async def generate(
...
-                best_response = response
-                logger.debug(
-                    "New best response:",
-                    data={"rating": best_rating, "response": best_response},
+                eval_response_attributes = (
+                    self.evaluator_llm.extract_response_message_attributes_for_tracing(
+                        response[0], "response"
+                    )
+                    if isinstance(response, list) and len(response) > 0
+                    else {}
```
saqadri: Same as before, don't quite understand why we're doing response[0]

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
```diff
@@ -158,116 +162,189 @@ async def generate(
...
-                iteration=refinement_count,
-            )
 
-            async with contextlib.AsyncExitStack() as stack:
-                if isinstance(self.optimizer, Agent):
-                    await stack.enter_async_context(self.optimizer)
+                logger.debug("Evaluator result:", data=evaluation_result)
+
+                # Track best response (using enum ordering)
+                if evaluation_result.rating.value > best_rating.value:
```
saqadri, line 260: should this add a span event in here?

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
```diff
@@ -158,116 +162,189 @@ async def generate(
...
+                        "New best response:",
+                        data={"rating": best_rating, "response": best_response},
+                    )
+
+                # Check if we've reached acceptable quality
+                if (
+                    evaluation_result.rating.value >= self.min_rating.value
+                    or not evaluation_result.needs_improvement
+                ):
+                    logger.debug(
```
saqadri, line 280: span event in here?

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
```diff
@@ -158,116 +162,189 @@ async def generate(
...
+                )
+
+                refinement_count += 1
 
-            refinement_count += 1
+            if isinstance(best_response, list) and len(best_response) > 0:
+                record_attributes(
+                    span,
+                    self.optimizer_llm.extract_response_message_attributes_for_tracing(
+                        best_response[0]
```
saqadri: Same as above, this looks odd [0]

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
```diff
@@ -276,15 +353,41 @@ async def generate_structured(
...
+                message=message, request_params=request_params
+            )
+
+            res = await self.optimizer.generate_structured(
+                message=response_str,
+                response_model=response_model,
+                request_params=request_params,
+            )
+
+            try:
```
saqadri: since the returned type should be a pydantic model, perhaps we should try `res.model_dump_json` here. I'm a bit worried about perf cost of this. I wish we could only do this if tracing was enabled

src/mcp_agent/tracing/tracer.py
rholinshead: I'll remove it and we can just use the global get_tracer for now

src/mcp_agent/workflows/embedding/embedding_openai.py
rholinshead: In cohere, `response.meta.tokens` is typed as `tokens: ApiMetaTokens | None`

But for openai here, the usage tokens aren't optional `prompt_tokens: int`

Are you saying that the typing is not necessarily correct?

src/mcp_agent/workflows/intent_classifier/intent_classifier_llm.py
```diff
@@ -102,40 +105,115 @@ async def create(
...
         self, request: str, top_k: int = 1
     ) -> List[LLMIntentClassificationResult]:
-        if not self.initialized:
-            await self.initialize()
+        tracer = self.context.tracer or trace.get_tracer("mcp-agent")
+        with tracer.start_as_current_span(
+            f"{self.__class__.__name__}.classify"
+        ) as span:
+            span.set_attribute("request", request)
+            span.set_attribute("intents", list(self.intents.keys()))
```
saqadri: I think this should be safe, just double check that self.intents will always exist

src/mcp_agent/workflows/llm/augmented_llm.py
```diff
@@ -188,6 +204,7 @@ def from_mcp_tool_result(
         """Convert an MCP tool result to an LLM input type"""
 
 
+# TODO: saqadri (MAC) - this likely needs to be converted to a Pydantic model
```
saqadri: Hmmm, this looks like old changes. Remove

src/mcp_agent/workflows/llm/augmented_llm.py
```diff
@@ -205,6 +222,7 @@ class AugmentedLLM(ContextDependent, AugmentedLLMProtocol[MessageParamT, Message
     def __init__(
         self,
         agent: Optional["Agent"] = None,
+        # TODO: saqadri (MAC) - Make this | None = None
```
saqadri: Hmmm, this looks like old changes. Remove

src/mcp_agent/workflows/llm/augmented_llm.py
```diff
@@ -284,29 +302,42 @@ async def select_model(
...
+        with tracer.start_as_current_span(
+            f"{self.__class__.__name__}.{self.name}.select_model"
+        ) as span:
+            span.set_attribute(GEN_AI_AGENT_NAME, self.agent.name)
+            model_preferences = self.model_preferences
+            if request_params is not None:
+                model_preferences = request_params.modelPreferences or model_preferences
+                model = request_params.model
+                if model:
+                    span.set_attribute("request_params.model", model)
```
saqadri, line 314: Should this use `AugmentedLLM.annotate_span_with_request_params(span, request_params)`

src/mcp_agent/workflows/llm/augmented_llm.py
```diff
@@ -459,6 +520,115 @@ def _log_chat_finished(self, model: Optional[str] = None):
...
+    def _extract_message_param_attributes_for_tracing(
+        self, message_param: MessageParamT, prefix: str = "message"
+    ) -> dict[str, Any]:
+        """
+        Return a flat dict of span attributes for a given MessageParamT.
+        Override this for the AugmentedLLM subclass MessageParamT type.
+        """
+        return {}
+
+    def _annotate_span_for_call_tool_result(
```
saqadri, line 601: This can probably be a static function

src/mcp_agent/workflows/llm/augmented_llm.py
```diff
@@ -459,6 +520,115 @@ def _log_chat_finished(self, model: Optional[str] = None):
         data = {"progress_action": "Finished", "model": model, "agent_name": self.name}
         self.logger.debug("Chat finished", data=data)
 
+    @staticmethod
```
saqadri, line 523: This and below are fine for now, but wondering if they should be split out into a utils as static methods. Not sure though

src/mcp_agent/workflows/llm/augmented_llm_azure.py
```diff
@@ -115,112 +128,153 @@ async def generate(self, message, request_params: RequestParams | None = None):
...
+            if params.use_history:
+                messages.extend(self.history.get())
 
-        if isinstance(message, str):
-            messages.append(UserMessage(content=message))
-        elif isinstance(message, list):
-            messages.extend(message)
-        else:
-            messages.append(message)
+            system_prompt = self.instruction or params.systemPrompt
```
saqadri, line 149: Should we log this

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
rholinshead: Thoughts on adding a `tracing_enabled` flag to context so that we can use it to check before doing expensive serialization for cases like this?

src/mcp_agent/mcp/mcp_aggregator.py
rholinshead: Good catch; same for arguments. Will add checks

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
rholinshead: Good catch. I think it was a mixup; will update to serialize each of the messages (behind tracing_enabled check)

src/mcp_agent/workflows/llm/augmented_llm.py
rholinshead: I think it _can_ but I think it will add a bunch of noise with all of the other params that are relevant for generation requests but not relevant here

src/mcp_agent/workflows/llm/augmented_llm.py
rholinshead: In recent changes I've added an initial check
```
if not self.context.tracing_enabled:
            return
```
So might be better to keep as-is now unless you feel strongly about making it static (and just doing the enabled check each time before calling it)

src/mcp_agent/workflows/llm/augmented_llm_azure.py
rholinshead: I can add it. Wasn't sure because it's used as the first message only if there are no other messages. Maybe I'll only add it as an attribute when it's used to prevent confusion

rholinshead:
Latest iteration includes changes from review:
- add `get_tracer` function to get the tracer from context
- add `tracing_enabled` flag to support checks to avoid unnecessary expensive serialization for tracing when not enabled
- annotate all messages in evaluator_optimizer instead of just the first
- remove old TODOs: also double-checked and compared against main for `agent.py`, `augmented_llm.py`, `augmented_llm_openai.py` and `augmented_llm_anthropic.py` to make sure no other stale changes are in
- added extra events noted in review
- re-tested each of the above examples with otel disabled and then enabled

src/mcp_agent/workflows/embedding/embedding_openai.py
saqadri: I'm worried about the usage property being null in some cases like when ollama or others are using the OpenAI api. Would be good to just check for null before indexing in just to be safe

src/mcp_agent/workflows/llm/augmented_llm_azure.py
```diff
@@ -354,10 +417,67 @@ def message_str(self, message: ResponseMessage) -> str:
...
+
+    def _extract_message_param_attributes_for_tracing(
+        self, message_param: MessageParam, prefix: str = "message"
+    ) -> dict[str, Any]:
+        """Return a flat dict of span attributes for a given MessageParam."""
+        attrs = {}
+        # TODO: rholinshead - serialize MessageParam dict
+        return attrs
+
+    def extract_response_message_attributes_for_tracing(
```
saqadri, line 474: Why are some of these functions private (prefixed with underscore but not this one?

src/mcp_agent/workflows/orchestrator/orchestrator.py
```diff
@@ -123,32 +125,91 @@ async def generate(
...
 
-        return [plan_result.result]
+            objective = str(message)
+            plan_result = await self.execute(objective=objective, request_params=params)
+
+            if self.context.tracing_enabled:
+                span.set_attribute("is_complete", plan_result.is_complete)
+                span.set_attribute("objective", plan_result.objective)
+                if plan_result.plan:
+                    span.set_attribute("plan.is_complete", plan_result.plan.is_complete)
```
saqadri: Duplicate?

src/mcp_agent/workflows/llm/augmented_llm_azure.py
rholinshead: The underscore ones are only being called by augmented llm instances; I originally had `_ extract_response_message_attributes_for_tracing` as well but then ended up needing to call it in a non-private/internal way in evaluator_optimizer, e.g. `self.evaluator_llm.extract_response_message_attributes_for_tracing`

PR #242, opened 2025-05-28
feat: Add additional OpenAI LLM config options

ttretau:
This adds additional, optional config options:
user - to add user argument to requests
default_headers - to add additional default headers

These are needed for example in custom, enterprise API environments.

## Summary by CodeRabbit

- **New Features**
	- Added support for specifying a user identifier and custom HTTP headers when configuring OpenAI model integrations. This enables enhanced customization and control over requests made to OpenAI services.

### DIFF
--- a/src/mcp_agent/config.py
+++ b/src/mcp_agent/config.py
@@ -159,6 +159,10 @@ class OpenAISettings(BaseModel):
 
     model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
 
+    user: str | None = None
+
+    default_headers: Dict[str, str] | None = None
+
 
 class AzureSettings(BaseModel):
     """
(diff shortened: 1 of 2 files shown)


src/mcp_agent/config.py
```diff
@@ -159,6 +159,10 @@ class OpenAISettings(BaseModel):
 
     model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
 
+    user: str | None = None
```
saqadri, line 162: Should this also be something we can set in RequestParams, or just in the config?

ttretau: Currently it looks to make sense on the config only.
Thanks for incorporating, @saqadri!
Looking forward for more collab..

PR #244, opened 2025-05-31
feat: Support mcp resources and prompts

StreetLamb:
## Summary by CodeRabbit

- **New Features**
  - Added resource management capabilities, enabling agents to list and access structured resources from connected servers.
  - Enhanced prompt creation by allowing resources to be used as context in agent applications.
  - Introduced comprehensive support for resource handling across OpenAI, Anthropic, Azure, Bedrock, and Google LLM integrations with dedicated converters.
  - Added utility functions for content extraction, MIME type detection, resource file handling, and URI normalization.
  - Provided a new example MCP server and agent demonstrating resource and prompt usage with setup and usage instructions.

- **Documentation**
  - Added detailed README and configuration files for the new MCP agent example.
  - Included setup instructions and example usage for connecting agents to custom MCP servers.

- **Bug Fixes**
  - Corrected minor documentation and docstring errors.

- **Tests**
  - Added extensive unit and integration tests for resource management, content utilities, MIME utilities, and multipart message conversions for all supported LLM providers.
  - Expanded LLM test coverage to verify handling of diverse input message types and structured output generation.


src/mcp_agent/agents/agent.py
```diff
@@ -455,6 +476,121 @@ def _annotate_span_for_tools_result(result: ListToolsResult):
...
+            )
+            return result
+
+    async def create_prompt(
+        self,
+        *,
+        prompt_name: str | None = None,
+        arguments: dict[str, str] | None = None,
+        resource_uris: list[str | AnyUrl] | str | AnyUrl | None = None,
+        server_name: str | None = None,
```
saqadri: It would be really great to be able to target multiple servers. For example, the resources may be from different servers, the prompt from another. Different resources could be from different servers, etc. I can update this post-merge

src/mcp_agent/utils/mime_utils.py
```diff
@@ -0,0 +1,69 @@
+# mime_utils.py
```
saqadri: nit: remove or add proper docstring

src/mcp_agent/mcp/mcp_aggregator.py
```diff
@@ -591,6 +637,151 @@ async def list_tools(self, server_name: str | None = None) -> ListToolsResult:
...
+                await self.load_servers()
+
+            span.set_attribute("uri", uri)
+
+            # If server_name is provided, use that server
+            if server_name:
+                span.set_attribute("server_name", server_name)
+            else:
+                # Use the URI to find the server name
+                server_name = self._find_server_name_from_uri(uri)
```
saqadri: nit: what do you think about updating `_parse_capability_name` to do _find_server_name_from_uri  when something like this gets called: `self._parse_capability_name(uri, "resource")`

src/mcp_agent/mcp/mcp_aggregator.py
```diff
@@ -990,6 +1181,29 @@ def getter(item: NamespacedPrompt):
         # No match found
         return None, None
 
+    def _find_server_name_from_uri(self, uri: str) -> str:
```
saqadri: Should we also enable some kind of uri convention for namespaced server uris, similar to how we namespace tool and prompt names? could be `server:{rest of uri}`. Not urgent, just curious if we should.

src/mcp_agent/workflows/llm/multipart_converter_anthropic.py
```diff
@@ -0,0 +1,518 @@
...
+from mcp_agent.utils.resource_utils import extract_title_from_uri
+from mcp_agent.workflows.llm.augmented_llm import MessageTypes
+
+_logger = get_logger("multipart_converter_anthropic")
+
+# List of image MIME types supported by Anthropic API
+SUPPORTED_IMAGE_MIME_TYPES = {"image/jpeg", "image/png", "image/gif", "image/webp"}
+
+
+class AnthropicConverter:
```
saqadri, line 50: For all these Converter classes -- 
Let's define a `Protocol`, similar to 

```
class ProviderToMCPConverter(Protocol, Generic[MessageParamT, MessageT]):
```

We can call it `MessageConverter` or something, and it can have the following APIs: `from_prompt_message_multipart`, `from_prompt_message`. Maybe also the inverse direction? But not sure if we need that.

StreetLamb: Will look into it in another PR. There are some similarities between these Converters and the current `ProviderToMCPConverter`, will see how we can converge them.

src/mcp_agent/mcp/mcp_aggregator.py
StreetLamb: Will work on this in another PR!

PR #252, opened 2025-06-05
Remove aexit call in mcp aggregator

kawolum:
I don't know if this is the best fix.

Temporal workflows when run with non local Temporal instance would get stuck at

1. shutdown_aggregator_task
2. self._persistent_connection_manager.__aexit__
3. self._tg.__aexit__

Then it would eventually error out with "Attempted to exit cancel scope in a different task than it was entered in"

## Summary by CodeRabbit

- **Refactor**
	- Improved the shutdown process by streamlining cleanup steps for connections.

### DIFF
--- a/src/mcp_agent/mcp/mcp_aggregator.py
+++ b/src/mcp_agent/mcp/mcp_aggregator.py
@@ -225,16 +225,6 @@ class MCPAggregator(ContextDependent):
                                         "Timeout during disconnect_all(), forcing shutdown"
                                     )
 
-                                # Ensure the exit method is called regardless
-                                try:
-                                    await self._persistent_connection_manager.__aexit__(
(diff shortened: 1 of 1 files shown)


saqadri:
Let me investigate a bit more, I've seen this error as well but haven't nailed down the root cause. Does it hang the temporal worker in your case?

kawolum:
Yes, the temporal worker would hang at self._tg.aexit then eventually get cancelled due to the activity timeout.
I was able to consistently see this problem.

saqadri:
Going to close this PR @kawolum -- I just pushed 71c8bb32ec122a4d3383bc2d32721ff3207f3021 for trying to make the exit more resilient. We can revisit the fix if this doesn't fix it

PR #264, opened 2025-06-11
Removing installation from local directory in example

andrew-lastmile:
## Summary by CodeRabbit

- **Chores**
  - Updated dependency reference for mcp-agent in requirements to use the package name instead of a local file path.

### DIFF
--- a/examples/mcp_agent_server/asyncio/requirements.txt
+++ b/examples/mcp_agent_server/asyncio/requirements.txt
@@ -1,6 +1,5 @@
 # Core framework dependency
-mcp-agent @ file://../../../  # Link to the local mcp-agent project root
-
+mcp-agent
 
 rich
 openai>=1.0.0


examples/mcp_agent_server/asyncio/requirements.txt
```diff
@@ -1,6 +1,5 @@
 # Core framework dependency
-mcp-agent @ file://../../../  # Link to the local mcp-agent project root
```
saqadri, line 2: Why this change? This is helpful to keep the examples in sync with the repo

andrew-lastmile: Good point. We can keep it as is.

For me, I ran into the issue where I was building this mcp agent server as a hostable docker image and since it had a reference to the local directory of mcp-agent. All the file paths for dependencies needed to go back to root and the additional libraries in the project root increased the image size by 30%.

saqadri: Ah interesting. Yeah we can have a separate requirements.txt and script for rolling this into a docker image perhaps

saqadri: Will close this PR for now

PR #265, opened 2025-06-11
Fix issues when running agents in threads

StreetLamb:
## Summary by CodeRabbit

- **New Features**
  - Introduced a multithreaded example demonstrating concurrent creative script generation and evaluation, including sample scripts, configuration, and requirements.
  - Added a sample server for counting words in scripts.

- **Bug Fixes**
  - Improved thread safety and resource management for logging, event bus, and server connections.
  - Enhanced cleanup and shutdown procedures for asynchronous and threaded operations.

- **Refactor**
  - Changed global context, registries, and event bus management to use thread-local storage and thread-safe singletons for better isolation and concurrency.
  - Updated workflow tasks for OpenAI, Azure, Google, Anthropic, and Ollama to use fully asynchronous clients with proper async resource management.
  - Refactored decorator, signal, and activity registries to thread-safe singleton patterns.
  - Modified connection manager to replace async synchronization primitives with thread-safe flags and locks, improving cross-thread safety.
  - Improved event loop handling in logging for better thread safety and fallback mechanisms.
  - Updated event bus to maintain isolated thread-local instances with lifecycle and cleanup management.
  - Adjusted persistent connection manager cleanup to avoid cross-context task group issues.

- **Documentation**
  - Added example configuration and secrets files for the multithreaded agent.

- **Tests**
  - Updated tests to align with new thread-safe shutdown state checks.
  - Improved test fixture to assign unique names dynamically for better isolation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,13 +18,13 @@ tests:
 
 .PHONY: coverage
 coverage:
-	uv run coverage run -m pytest
+	uv run coverage run -m pytest tests
 	uv run coverage xml -o coverage.xml
 	uv run coverage report -m --fail-under=80
 
 .PHONY: coverage-report
 coverage-report:
-	uv run coverage run -m pytest
+	uv run coverage run -m pytest tests
 	uv run coverage html
 
 .PHONY: schema


src/mcp_agent/app.py
```diff
@@ -89,9 +91,9 @@ def __init__(
         # We initialize the task and decorator registries at construction time
         # (prior to initializing the context) to ensure that they are available
         # for any decorators that are applied to the workflow or task methods.
-        self._task_registry = ActivityRegistry()
-        self._decorator_registry = DecoratorRegistry()
-        self._signal_registry = SignalRegistry()
+        self._task_registry = get_global_activity_registry()
+        self._decorator_registry = get_global_decorator_registry()
+        self._signal_registry = get_global_signal_registry()
```
saqadri, line 98: Do we need to make these global? I'd prefer to keep them instantiated per MCPApp, though if there's a specific reason for this perhaps we can discuss

src/mcp_agent/mcp/mcp_connection_manager.py
```diff
@@ -111,23 +137,28 @@ async def initialize_session(self) -> None:
...
+        self._set_initialized_flag(True)
         self._initialized_event.set()
 
     async def wait_for_initialized(self) -> None:
         """
-        Wait until the session is fully initialized.
+        Wait until the session is fully initialized using an async event with timeout.
         """
-        await self._initialized_event.wait()
+        with move_on_after(30):
```
saqadri, line 156: Should we wait just 30s? Or should this be perpetual until the flag is set?

saqadri (commented):
Noticed a few things that I left comments on. Will test this in the branch myself as well. Overall great work, this is a very nontrivial change. Just need to look at it a bit more carefully before merging in

src/mcp_agent/mcp/mcp_connection_manager.py
StreetLamb: I think its better to be a fixed duration so it doesn't hang indefinitely if the server fails to initialise.

src/mcp_agent/app.py
StreetLamb: I made it global because it was giving me problems on Temporal as the workers and app have different registries

saqadri:
@StreetLamb I did a bunch of research and will prefer to close this in favor of #367, which includes some of the examples you added in this change :)

PR #272, opened 2025-06-12
fix: Add user request param and missing default_headers

ttretau:
This adds in addition to 6799107f9 the user as option for request params [^1]. Also fixing the missing option for default_headers in structured completion.

[^1]: https://platform.openai.com/docs/guides/safety-best-practices#end-user-ids


ttretau:
> Great addition! However, don't we need to use the `user` param in OpenAIAugmentedLLM (passing it into the request)?
> 
> Might also be good to log in the trace span as an attribute

Thanks for the input, @saqadri ! 
I hope it got setup now correct with different options. Added also some test cases.

PR #276, opened 2025-06-16
TemporalExecutor.start_workflow workflow_id rethink

jsedgwick:
This is the safe, backwards compatible version of this change, but if you look at the examples, where the preferred route looks very much like what we will do automatically for users if no workflow_id is provided - {WorkflowClassName}-{uuid} - you can definitely make the case that we should just accept that as a fallback instead of warning about it.

## Summary by CodeRabbit

- **New Features**
	- Each workflow execution now uses an explicit, unique workflow ID for better identification and tracking.

- **Bug Fixes**
	- Improved handling of workflow ID assignment to avoid potential conflicts in workflow executions.

- **Documentation**
	- Enhanced usage instructions and examples for starting workflows, including updated docstrings.

- **Tests**
	- Expanded test coverage for workflow ID handling, argument passing, and result waiting in workflow execution.

- **Chores**
	- Removed a legacy test suite related to conversation management.


jsedgwick:
Implemented in https://github.com/lastmile-ai/mcp-agent/commit/c30c826f6e1d9882f1cec2847f2bd8c36831964c

PR #279, opened 2025-06-22
Add support for elicitation

StreetLamb:
Adding support for elicitation feature which has been merged into the MCP python-sdk but is not yet included in a released version.

## Summary by CodeRabbit

* **New Features**
  * Introduced elicitation support, enabling agents to pause and request structured user input or confirmation during tool execution.
  * Added a console-based elicitation handler with slash command support for interactive input and control.
  * Expanded callback handler options in the application to include elicitation callbacks.
  * Provided example scripts, server, configuration, and documentation for elicitation workflows.

* **Improvements**
  * Enhanced human input handler with slash command support for declining, cancelling, or requesting help during input.
  * Updated agent client session to process elicitation requests and handle user responses or interruptions.

* **Documentation**
  * Added comprehensive README and example configuration files to guide setup and usage of elicitation features.


src/mcp_agent/human_input/handler.py
saqadri, line 1: Overall this looks solid, though I'm wondering if we should keep elicitation separate so users can use it separately or as desired.  Concretely it would look very similar to this but be an `elicitation` folder similar to `human_input`. 

In the near future we can consider merging the concepts into one.

What do you think @StreetLamb?

saqadri (approved):
One main comment and also a failing test gh workflow, otherwise change looks good. Approving to unblock!

saqadri:
Amazing work!

PR #283, opened 2025-06-25
Fix for workflow signals in temporal

romanvanderkrogt:
Discovered two bugs while trying to get human interaction to work from a temporal workflow:

- the HumanInputRequest should include the workflow_id and run_id, so that when the input is provided, the signal can be sent to the correct workflow
- temporal workflows cannot signal themselves; instead the signal handler can directly add deliver the signal to the mailbox

Also added a helper `InteractiveWorkflow` class that exposes the information asked for by the LLM through a `query`, and a `signal` to provide the requested input. A new example demonstrates how to use this workflow.

## Summary by CodeRabbit

- **New Features**
  - Introduced an interactive Temporal workflow example demonstrating human-in-the-loop agent workflows with UI-based human input and response handling.
  - Added an `InteractiveWorkflow` class enabling workflows to pause and await user input, featuring query and signal methods for interaction.
- **Enhancements**
  - Enhanced human input request handling by including workflow and run identifiers for precise workflow context association.
  - Standardized human input request creation across executor implementations.
- **Bug Fixes**
  - Prevented workflows from signaling themselves in Temporal to avoid runtime errors.
- **Documentation**
  - Provided usage instructions and example code for running and interacting with the new interactive workflow.

### DIFF
--- a/src/mcp_agent/executor/temporal/interactive_workflow.py
+++ b/src/mcp_agent/executor/temporal/interactive_workflow.py
@@ -1,0 +1,84 @@
+import asyncio
+from dataclasses import dataclass
+from typing import Generic, TypeVar
+
+from mcp_agent.executor.workflow import Workflow
+from mcp_agent.human_input.types import HumanInputRequest, HumanInputResponse
+from mcp_agent.logging.logger import get_logger
+
+from temporalio import workflow
+
+logger = get_logger(__name__)
+
+T = TypeVar("T")
+
+
+@dataclass
+class HumanResponse:
+    response: str
+
+
+class InteractiveWorkflow(Workflow[T], Generic[T]):
+    """
+    A workflow with support for handling human input requests and responses.
+
+    Example:
+        To use this workflow, create a workflow like this:
+
+        @app.workflow
+        class MyWorkflow(InteractiveWorkflow):
+            @app.workflow_run
+            async def run(self, input: str) -> WorkflowResult[str]:
+                interactive_agent = Agent(
+                    name="basic_interactive_agent",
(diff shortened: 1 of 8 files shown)


src/mcp_agent/executor/temporal/interactive_workflow.py
```diff
@@ -0,0 +1,84 @@
...
+                # etc.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._lock = asyncio.Lock()
+        self._request: HumanInputRequest = None
+        self._response: str = None
+
+    @workflow.query
```
saqadri, line 47: aside: @romanvanderkrogt what do you think about exposing this via the MCPApp decorators? Do you think there are instances where query would be useful beyond this case?

saqadri (approved):
Superb change @romanvanderkrogt, thanks for contributing. I'll accept it as-is -- great work!

src/mcp_agent/executor/temporal/interactive_workflow.py
romanvanderkrogt: Yes, for long running workflows (like deep research-style ones) I'm going to need a way to show the user the progress, for example. A query would be a great way to expose that in the case of temporal. I suppose it depends on your plans for the asyncio executor. Do you envision people running that as a server or is it more of a cli alternative? If it's the latter, there's no need for queries in the asyncio case, I think. It would just add unnecessary complexity and overhead. (Tbh, if one would want to run it as a server, they should go with temporal anyway, but that's just my opinion 😄)

PR #291, opened 2025-06-30
[bugfix] Fix errors in temporal workflow app server

rholinshead:
## Summary
Fix a few errors that I noticed when running the `mcp_agent_server/temporal` example.

Mainly:

### Fix Run Response Type Annotation
Previously, was seeing this error when trying to run a workflow:
```
[ERROR] 2025-06-30T14:36:22 
mcp_agent.workflow_mcp_client - 
Workflow run Error executing tool 
workflows-BasicAgentWorkflow-run: 1 
validation error for runOutput
result
  Input should be a valid dictionary 
[type=dict_type, 
input_value='0197c220-3d08-78ba-b42a-6
429be46de11', input_type=str]
    For further information visit 
https://errors.pydantic.dev/2.11/v/dic
t_type status is empty. 
get_status_result:
```

The issue was that FastMCP was trying to serialize the result as `Dict[str, Any]` even though it's actually a string. Fixing the annotation resolves this error.

### Fix Internal Temporal Status Call
Even with workflows succeeding, attempting to get their status would have an error in the temporal data:
```
"temporal": {
          "id": "BasicAgentWorkflow",
          "workflow_id": 
"BasicAgentWorkflow",
          "run_id": 
"0197c236-464c-7324-a071-be93340f4d67"
,
          "status": "ERROR",
          "error": "workflow execution
not found for workflow ID 
\"BasicAgentWorkflow\" and run ID 
\"0197c236-464c-7324-a071-be93340f4d67
\""
        }
      }
    },
```

This is because the workflow ID is being passed through with the workflow name. Temporal expects the ID to be the actual workflow ID (which we dynamically generate a new unique ID when executing a workflow) so it never finds it by the Workflow Definition name.

To fix, update the workflow_id passed in to be the actual id instead of name, when possible.

## Testing
`uv run pytest`

Ran the `mcp_agent_server/temporal` without any errors now.

Also ran `mcp_agent_server/asyncio` without any errors

## Summary by CodeRabbit

* **New Features**
  * Workflow execution results now include both workflow ID and run ID, providing clearer identification for each workflow run.

* **Bug Fixes**
  * Improved consistency in workflow status retrieval by preferring workflow IDs over names where available.

* **Documentation**
  * Updated documentation to reflect changes in workflow execution return types and usage.

* **Tests**
  * Adjusted test cases to validate the new workflow execution result structure. 

* **Chores**
  * Updated client examples to display both workflow ID and run ID after starting a workflow.


src/mcp_agent/server/app_server.py
```diff
@@ -341,7 +341,7 @@ def create_workflow_specific_tools(
     async def run(
         ctx: MCPContext,
         run_parameters: Dict[str, Any] | None = None,
-    ) -> Dict[str, Any]:
```
saqadri, line 344: Hmm on second thought we should probably return both workflow ID and run ID now, since both are unique identifiers

saqadri: Maybe we return a typeddict or dataclass

saqadri (approved):
Great catches. I think we should return workflow if as well from the run endpoint. Otherwise looks great. Can also fix on top

src/mcp_agent/server/app_server.py
rholinshead: Looks like the mcp layer can't serialize a dataclass via FastTool, so keeping as a Dict[str, str] there. Updating run_async to return a `WorkflowExecution` type with both IDs

PR #297, opened 2025-07-04
Add support for claude models via VertexAI and Bedrock in AnthropicAugmentedLLM

StreetLamb:


src/mcp_agent/workflows/llm/augmented_llm_anthropic.py
```diff
@@ -114,11 +133,17 @@ def __init__(self, *args, **kwargs):
             intelligencePriority=0.3,
         )
 
-        default_model = "claude-3-7-sonnet-latest"  # Fallback default
+        default_model = "claude-3-7-sonnet-20250219"
```
saqadri: Should this model id just be "claude-sonnet-4-0"?

https://docs.anthropic.com/en/docs/about-claude/models/overview

In general we should update the defaults to 4.0

saqadri (approved):
Awesome work! Just one request to update default model ids and use model aliases for default case

src/mcp_agent/config.py
```diff
@@ -133,6 +131,19 @@ class BedrockSettings(BaseModel):
...
 
+class AnthropicSettings(VertexAISettings, BedrockSettings):
+    """
+    Settings for using Anthropic models in the MCP Agent application.
+    """
+
+    api_key: str | None = None
+    default_model: str | None = None
+    bedrock: bool = False
+    vertexai: bool = False
```
saqadri: What about making this an enum or literal union type (similar to `execution_engine`?

I was thinking something like `provider: "anthropic" | "bedrock" | "vertexai"`

PR #299, opened 2025-07-04
Support loading Settings from a literal value via MCP_APP_SETTINGS_PRELOAD

petersonbill64:
Meant to support secure production cases where the config may contain literal secrets, so the preferred method of loading is with a process-scoped environment variable like `MCP_APP_SETTINGS_PRELOAD="$(load_the_value ...)" uv run main.py`.

This is a common pattern in production container deployments, because it avoids writing plaintext secrets to a file in the container and avoids setting a container-level environment variable in plaintext, both of which can be inspected externally.

## Summary by CodeRabbit

* **New Features**
  * Support for loading app configuration from a YAML string provided via environment variables, with optional strict parsing.
* **Bug Fixes**
  * Improved handling of invalid preload configuration (strict mode raises errors; lenient mode falls back safely).
* **Tests**
  * Added tests covering preload behavior, override semantics, and strict/lenient error handling.
* **Chores**
  * Added several runtime dependencies (including pydantic-yaml) and updated .gitignore to ignore IDE files.


.idea/workspace.xml
saqadri, line 1: Is this meant to be merged in? If not, would recommend adding to `.gitignore`:

```
.idea/workspace.xml
.idea/*_workspace.xml
```

saqadri (approved):
Ship it. Just need to run `make schema` to regenerate the config schema and add that to commit

Issue #302, opened 2025-07-07
Bedrock request error when multiple MCP tools are called


hayashikun:
When using the MCP agent with Bedrock and multiple tools are called in a single request, following error occurs:

```
[mcp_agent.executor.executor] Error executing task: An error occurred (ValidationException) when calling the Converse operation: Expected toolResult blocks at messages.2.content for the following Ids: tooluse_9ySdT-H2QuWYlqrScSBHuA, tooluse_JKkvVfhuTOGhrwnHY0lLQQ
[mcp_agent.workflows.llm.augmented_llm_bedrock.demo_agent] Error: An error occurred (ValidationException) when calling the Converse operation: Expected toolResult blocks at messages.2.content for the following Ids: tooluse_9ySdT-H2QuWYlqrScSBHuA, tooluse_JKkvVfhuTOGhrwnHY0lLQQ
```

Following messages are sent to the Bedrock LLM:
```
[
  {'content': [{'text': 'Generate 3 different random numbers'}], 'role': 'user'},
  {'content': [
    {'text': 'I'll generate 3 different random numbers for you using the random number generator.'},
    {'toolUse': {'input': {}, 'name': 'random-number_rand', 'toolUseId': 'tooluse_EhfEmT1xQeW8BIS-ZAwCNQ'}},
    {'toolUse': {'input': {}, 'name': 'random-number_rand', 'toolUseId': 'tooluse_9ySdT-H2QuWYlqrScSBHuA'}},
    {'toolUse': {'input': {}, 'name': 'random-number_rand', 'toolUseId': 'tooluse_JKkvVfhuTOGhrwnHY0lLQQ'}}
  ], 'role': 'assistant'},
  {'content': [{'toolResult': {'content': [{'text': '92'}], 'status': 'success', 'toolUseId': 'tooluse_EhfEmT1xQeW8BIS-ZAwCNQ'}}], 'role': 'user'},
  {'content': [{'toolResult': {'content': [{'text': '56'}], 'status': 'success', 'toolUseId': 'tooluse_9ySdT-H2QuWYlqrScSBHuA'}}], 'role': 'user'},
  {'content': [{'toolResult': {'content': [{'text': '55'}], 'status': 'success', 'toolUseId': 'tooluse_JKkvVfhuTOGhrwnHY0lLQQ'}}], 'role': 'user'}
]
```

This is because the Bedrock LLM expects tool results to be returned in the next message after the tool uses.

[...]

PR #303, opened 2025-07-07
fix: combine multiple tool results into single message in Bedrock LLM

hayashikun:
close https://github.com/lastmile-ai/mcp-agent/issues/302

Previously, when handling multiple tool uses in a single response, each tool result was incorrectly added as a separate user message. This fix collects all tool results and combines them into a single user message, matching the expected Bedrock API format.

Changes:
- Modified tool result handling in augmented_llm_bedrock.py to batch results
- Added test_multiple_tool_usage to verify the fix
- Added create_multiple_tool_use_response helper for testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of multiple tool uses in responses, consolidating results into a single message for clearer output.

* **Tests**
  * Added new tests to verify correct processing of responses containing multiple tool uses in one message.


saqadri (approved):
Awesome fix @hayashikun, thank you for the contribution!!

PR #306, opened 2025-07-09
Fix/tool arguments prefix

rholinshead:
## Description
Fixing the mcp tracing example to have arguments annotated under `arguments` prefix, similar to other places in the codebase. When testing, found and fixed a couple issues:
- example requirements.txt path to mcp-agent package root has one too many `../` now. Remove
- if the server is not running there is a `ServerInitializationError` that would result in an error in annotating the span; it turns out the CallToolResult type is just a suggestion in that case and the `ServerInitializationError` has no `isError` attribute. So, fix to do proper `attr` checks instead of indexing by attribute name. Pulled into a helper for reuse

## Testing
- Ran the tracing example successfully now; with server not running ensure there is no longer an error thrown in annotating the span

Correct span args:
Screenshot 2025-07-09 at 1 44 51 PM
Screenshot 2025-07-09 at 3 16 12 PM
Screenshot 2025-07-09 at 3 16 24 PM

## Summary by CodeRabbit

* **New Features**
  * The "magic number" tool now accepts a user-provided input number and returns the sum with 42.
  * Tool descriptions and input requirements are more explicit and user-friendly.

* **Bug Fixes**
  * Test cases updated to align with the enhanced tool input and output behavior.

* **Chores**
  * Updated local agent package dependency path.
  * Enhanced tracing and error reporting for tool calls with improved span annotations.

### DIFF
--- a/examples/tracing/mcp/main.py
+++ b/examples/tracing/mcp/main.py
@@ -20,13 +20,16 @@ async def test_sse():
             server_names=["mcp_test_server_sse"],
         )
 
+        original_number = 1
+
         async with agent:
             print(await agent.list_tools())
             call_tool_result: CallToolResult = await agent.call_tool(
-                "mcp_test_server_sse_get-magic-number"
+                "mcp_test_server_sse_get-magic-number",
+                {"original_number": original_number},
             )
 
-            assert call_tool_result.content[0].text == "42"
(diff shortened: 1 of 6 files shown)


saqadri (approved):
Awesome!

PR #330, opened 2025-07-24
Fix empty mcp tool result error

### LINKED ISSUE #328: Fix OpenAIAugmentedLLM to handle tool output of empty list or None
Reproduction steps:

1. Have an MCP tool that returns [] or None

```
@mcp.tool()
def return_nothing():
    """return Nothing
    """
    return []

@mcp.tool()
def return_nothing2():
    """return Nothing2
    """
    return None
```
2. Have an agent use those tools
3. The agent would fail with 
```
Error executing task: Error serving response: got non-ok status code 400 body: {
  "error": {
    "message": "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following 
tool_call_ids did not have response messages: call_12345",
    "type": "invalid_request_error",
    "param": "messages",
    "code": null
  }
}
```

What happens:

1. When those tools are called, it returns a result with empty content list.
2. When it gets [here](https://github.com/lastmile-ai/mcp-agent/blob/7d254fa1db52038b9feef91723e248c1b31fa677/src/mcp_agent/workflows/llm/augmented_llm_openai.py#L536), it doesn't create the tool completion message and then it fails later with the above message. 
    1. Not sure if this is a problem for other LLMs
    2. Not sure if this is the right place for the fix, maybe it's not handling the response correctly beforehand

Context:

1. The tool may return None or [] if it's some data querying tool and there is no result.
2. Returning {} works surprisingly.

StreetLamb:
Fixes #328 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency in tool call execution, ensuring a response is always returned.

### DIFF
--- a/src/mcp_agent/mcp/mcp_aggregator.py
+++ b/src/mcp_agent/mcp/mcp_aggregator.py
@@ -871,9 +871,6 @@ class MCPAggregator(ContextDependent):
                     res = await client.call_tool(
                         name=local_tool_name, arguments=arguments
                     )
-                    if len(res.content) == 0:
-                        # Treat none or empty list from tool result as tool result
-                        res.content = [TextContent(type="text", text="[]")]
                     _annotate_span_for_result(res)
                     return res
                 except Exception as e:
--- a/src/mcp_agent/mcp/mcp_aggregator.py
+++ b/src/mcp_agent/mcp/mcp_aggregator.py
@@ -871,6 +871,9 @@ class MCPAggregator(ContextDependent):
                     res = await client.call_tool(
                         name=local_tool_name, arguments=arguments
                     )
+                    if len(res.content) == 0:
+                        # Treat none or empty list from tool result as tool result
+                        res.content = [TextContent(type="text", text="[]")]
                     _annotate_span_for_result(res)
                     return res
                 except Exception as e:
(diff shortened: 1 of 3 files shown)


src/mcp_agent/mcp/mcp_aggregator.py
```diff
@@ -871,6 +871,9 @@ async def try_call_tool(client: ClientSession):
                     res = await client.call_tool(
                         name=local_tool_name, arguments=arguments
                     )
+                    if len(res.content) == 0:
+                        # Treat none or empty list from tool result as tool result
+                        res.content = [TextContent(type="text", text="[]")]
```
rholinshead: It seems off to require adding a placeholder TextContent when the types suggest a `CallToolResult` with empty content is valid.

Do you know if this issue happens for other LLMs? My understanding is that the error means there is a subsequent call to the LLM with all of the messages (including original tool call), and that's missing any tool call response message.

Looking through the code I suspect it might be here: https://github.com/lastmile-ai/mcp-agent/blob/7d254fa1db52038b9feef91723e248c1b31fa677/src/mcp_agent/workflows/llm/augmented_llm_openai.py#L536-L545

So, we probably want to leave the res.content as empty and then just handle the empty result there (maybe empty string or "No result returned" similar to other classes implementing a `from_mcp_tool_result` function (side note -- I don't think openai augmented llm has its own `from_mcp_tool_result`, if that matters)

rholinshead (approved):
Nice, thanks!

PR #331, opened 2025-07-24
Maybe Fix: Use lock when getting aggregator for agent

rholinshead:
## Description
Have received a report of spurious errors in the temporal worker when running workflows:
```
{
  "message": "Server aggregrator for agent 'Finder' not found",
  "stackTrace": "...dist/lib/python3.11/site-packages/mcp_agent/agents/agent.py\", line 1155, in call_tool_task
    raise ValueError(f\"Server aggregrator for agent '{agent_name}' not found\")
",
  "applicationFailureInfo": {
    "type": "ValueError"
  }
}
```

 when it happens, cancelling the workflow and retrying has worked.
I haven't been able to repro but looking through the code I do see that we are relying on obtaining an aggregator from a Dict that is managed with a lock, but not using a lock for accessing the value from the Dict. Sort of a Hail Mary attempt here with just ensuring the access is also wrapped with the lock to prevent a potential race condition it trying to get the aggregator before it's initialized for the agent.

## Test Plan
```
make lint
make tests
```

Run each of the examples under `examples/temporal` (with fix for router example, which was erroring on not passing input str)

## Summary by CodeRabbit

* **Refactor**
  * Simplified workflow invocation in the example Temporal router by removing the need to provide input parameters.
  * Centralized and standardized the retrieval of server aggregators in agent tasks for improved consistency and error handling.

### DIFF
--- a/examples/temporal/router.py
+++ b/examples/temporal/router.py
@@ -38,12 +38,12 @@ class RouterWorkflow(Workflow[str]):
     """
 
     @app.workflow_run
-    async def run(self, input: str) -> WorkflowResult[str]:
+    async def run(self) -> WorkflowResult[str]:
         """
-        Run the workflow, processing the input data.
+        Run the workflow.
 
         Args:
-            input_data: The data to process
+            N/A
 
         Returns:
             A WorkflowResult containing the processed data
@@ -150,7 +150,6 @@ async def main():
 
         handle = await executor.start_workflow(
             "RouterWorkflow",
-            None,
         )
         a = await handle.result()
         print(a)


StreetLamb:
I am able to replicate the error by running multiple temporal workers in parallel. My suspicion of whats happening:
- Two workers A and B running.
- During agent creation part of the workflow, Agent A picks up the `initialize_aggregator_task` activity and stores the aggregator in its `server_aggregators_for_agent` dict. 
- During tool call part of the workflow, Worker B picks up the `call_tool_task` activity instead which fails as its `server_aggregators_for_agent` dict is empty.

Maybe one solution is to store the `server_aggregators_for_agent` dict in context instead of in `AgentTasks` class? This way if any worker picks it up it can retrieve the agent's aggregator from context.

rholinshead:
> I am able to replicate the error by running multiple temporal workers in parallel. My suspicion of whats happening:
> 
> * Two workers A and B running.
> * During agent creation part of the workflow, Agent A picks up the `initialize_aggregator_task` activity and stores the aggregator in its `server_aggregators_for_agent` dict.
> * During tool call part of the workflow, Worker B picks up the `call_tool_task` activity instead which fails as its `server_aggregators_for_agent` dict is empty.
> 
> Maybe one solution is to store the `server_aggregators_for_agent` dict in context instead of in `AgentTasks` class? This way if any worker picks it up it can retrieve the agent's aggregator from context.

Isn't the context going to be distinct for each worker, though?

There's also a similar issue being seen where `initialize_aggregator_task` called multiple times for the same agent name but different server names is returning initialize response for the original server names. Though I'm not sure whether this is an expected use case for an agent (i.e. expect either all servers are given to the agent or use different agent names for different server subsets)?

saqadri:
@kawolum I am pretty sure I found the root cause for the temporal concurrency issue. It's been fixed in #367 -- very subtle initialization of the lock. I will submit a patch update over the weekend so you can try it out.

kawolum:
> @kawolum I am pretty sure I found the root cause for the temporal concurrency issue. It's been fixed in #367 -- very subtle initialization of the lock. I will submit a patch update over the weekend so you can try it out.

Let me know when this is released, I can consume the update.

saqadri:
> > @kawolum I am pretty sure I found the root cause for the temporal concurrency issue. It's been fixed in #367 -- very subtle initialization of the lock. I will submit a patch update over the weekend so you can try it out.
> 
> 
> 
> Let me know when this is released, I can consume the update.

@kawolum v0.1.11 is published please pull and lmk if you have any issues

PR #366, opened 2025-08-08
Feat/temporal rpc metadata

rholinshead:
## Description
Support propagating rpc-metadata through temporal client requests. This is useful for setting request headers for the rpc requests. In our case, we would like to be able to propagate an api token through the request but Temporal only supports sending the api token parameter across when TLS is enabled. We're running the app and worker on local internal container network without TLS and need the api token passed through. We can leverage Authorization headers to pass it as the Bearer token using this PR change.

## Testing
Run the temporal example for basic agent. Dump the request traffic to a file
```
sudo tcpdump -i lo0 -s 0 -w temporal.pcap port 7233
```
and sniff it with Wireshark to see the headers set in the GRPC request:
Screenshot 2025-08-08 at 5 58 19 PM

### DIFF
--- a/LLMS.txt
+++ b/LLMS.txt
@@ -1536,6 +1536,7 @@ Example usage:
   - `max_concurrent_activities` (int | None) = None
   - `api_key` (str | None) = None
   - `timeout_seconds` (int | None) = 60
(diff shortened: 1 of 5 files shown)


saqadri (approved):
Looks great, ship it!

Issue #375, opened 2025-08-13
feat: add support for observability dashboard like LangFuse or LangSmith


alessandrocornacchia:
Would be nice to have out-of-the-box support for LLM-specific observability platforms.

One that I particularly like is [LangFuse](https://langfuse.com), which has built-in functionalities to score and evaluate reasoning trajectories and agent performance, e.g., using LLM-as-a-judge.

Jaeger is basically for tracing only, which is fine but is only one part of the story.

What do the folks think?

saqadri:
> Would be nice to have out-of-the-box support for LLM-specific observability platforms.
> 
> One that I particularly like is [LangFuse](https://langfuse.com), which has built-in functionalities to score and evaluate reasoning trajectories and agent performance, e.g., using LLM-as-a-judge.
> 
> Jaeger is basically for tracing only, which is fine but is only one part of the story.
> 
> What do the folks think?

@alessandrocornacchia thank you for the request and I completely agree. I really like Langfuse as well. With our otel tracing set up and the use of the [LLM semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) I think this should be possible to do with low effort.

Do you want to contribute this? I will be happy to collaborate on it! Also cc @rholinshead who added the otel tracing in the framework.

rholinshead:
I did a quick read of the Langfuse docs and it looks like they support traces in a number of different ways. What I think might work best is if we just support exporting to Langfuse in the existing OpenTelemetry format instead of adding langfuse-specific dependencies. I took a quick stab at that in https://github.com/lastmile-ai/mcp-agent/pull/376. I see that it tracks costs already, but there may be some additional work to ensure the trace data includes what Langfuse expects for formulating the data necessary for some of the other built-in functionalities (e.g. I think input/output null without one of the expected attributes -- https://langfuse.com/integrations/native/opentelemetry#trace-level-attributes), if you are able to test it out and see how it works for your use cases?

taqtiqa-mark:
@saqadri there is also the UI example of https://github.com/rinadelph/Agent-MCP which might provide some inspiration:

<img width="1426" height="814" alt="Image" src="https://github.com/user-attachments/assets/9d2603b2-0992-49db-bf44-d272f274c976" />
<img width="1426" height="814" alt="Image" src="https://github.com/user-attachments/assets/2e9ccd56-16f5-43dd-b163-0bcac6306fa7" />
[...]

PR #376, opened 2025-08-13
[Feature] Support OTLP exporter headers

rholinshead:
# Description
As mentioned in #375 it would be nice to support integration with observability tools like Langfuse via tracing. Langfuse makes this super easy by supporting exporting OpenTelemetry traces to a Langfuse otlp endpoint and then handling any necessary massaging of traces to the Langfuse format (https://langfuse.com/integrations/native/opentelemetry). This exporter requires Basic Authorization headers to be used for the otlp export requests, so this PR adds support for headers in the `OTLPSettings` in `mcp-agent`.

## Testing
Created new example for Langfuse trace exporting and can see these traces in Langfuse:

Screenshot 2025-08-13 at 2 05 21 PM

Note that trace doesn't have input/output; its root span is the agent initialize (KP with tracer being closed before the root span gets written)

## Summary by CodeRabbit

- **New Features**
  - Langfuse tracing example demonstrating agent tracing, LLM switching, and OTLP export with configurable headers (e.g., Authorization).
  - OTLP exporter accepts an optional endpoint and custom HTTP headers.

- **Documentation**
  - Added a README with step-by-step setup, secrets/Authorization guidance, and run instructions for the Langfuse example.

- **Chores**
  - Included ready-to-run configs, secrets template, and example-specific dependencies for local execution.

### DIFF
--- a/src/mcp_agent/config.py
+++ b/src/mcp_agent/config.py
@@ -268,8 +268,11 @@ class TraceOTLPSettings(BaseModel):
     Settings for OTLP exporter in OpenTelemetry.
     """
 
-    endpoint: str
+    endpoint: str | None = None
     """OTLP endpoint for exporting traces."""
+
+    headers: Dict[str, str] | None = None
+    """Optional headers for OTLP exporter."""
 
 
 class OpenTelemetrySettings(BaseModel):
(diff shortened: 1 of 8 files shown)


src/mcp_agent/config.py
```diff
@@ -268,9 +268,12 @@ class TraceOTLPSettings(BaseModel):
     Settings for OTLP exporter in OpenTelemetry.
     """
 
-    endpoint: str
+    endpoint: str | None = None
```
rholinshead, line 415: Made this optional so the secrets config is valid when it defines only the auth headers (instead of duplicating the endpoint)

saqadri (approved):
Love it!!! Thanks for adding this so quickly

PR #377, opened 2025-08-13
Allow a custom MCP server definition with MCPApp

saqadri:
The overall idea is to let a developer define their own `FastMCP` server definition, in order to define custom tools they may want on their agent server.

`MCPApp` now optionally takes the `FastMCP` object as part of its constructor arguments.

`create_mcp_server_for_app` now handles both cases:

* If app.mcp is provided, we attach the app to the server (mcp.mcp_app = app) and register per-workflow tools immediately so they show up in list_tools without relying on our lifespan.
* If we instantiate FastMCP, we continue to use our lifespan; per-workflow tools are registered there as before.

Tool handlers now prefer the attached app for workflow discovery and registry access when there’s no `ServerContext`:
* workflows-list reads from ctx.fastmcp.mcp_app.workflows when no `ServerContext`.
* workflows-run/get_status fallback to the attached app’s context/workflow_registry if lifespan_context isn’t our `ServerContext`.

## Summary by CodeRabbit

- New Features
  - Added a token-usage tool to report workflow/run input/output/total tokens, estimated cost, per-model breakdown, and a usage tree.
  - App can now attach to an existing MCP server, enabling bring-your-own-server setups.
- Refactor
  - Server logic updated to support both managed and externally managed startup modes seamlessly.
- Documentation
  - Constructor docs updated to describe the new server integration option.

### DIFF
--- a/src/mcp_agent/server/app_server.py
+++ b/src/mcp_agent/server/app_server.py
@@ -6,7 +6,7 @@ mcp-agent workflows and agents as MCP tools.
 import json
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
-from typing import Any, Dict, List, Type, TYPE_CHECKING
+from typing import Any, Dict, List, Optional, Tuple, Type, TYPE_CHECKING
 
 from mcp.server.fastmcp import Context as MCPContext, FastMCP
 from mcp.server.fastmcp.exceptions import ToolError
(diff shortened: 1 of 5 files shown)


src/mcp_agent/server/app_server.py
```diff
@@ -112,16 +159,35 @@ async def app_specific_lifespan(mcp: FastMCP) -> AsyncIterator[ServerContext]:
...
+            # TODO: saqadri (MAC) - create a much more detailed description
+            # based on all the available agents and workflows,
+            # or use the MCPApp's description if available.
+            instructions=f"MCP server exposing {app.name} workflows and agents. Description: {app.description}",
+            lifespan=app_specific_lifespan,
+            **kwargs,
+        )
+        # Store the server on the app so it's discoverable and can be extended further
+        app.mcp = mcp
+        setattr(mcp, "_mcp_agent_app", app)
```
rholinshead, line 190: Is `_mcp_agent_server_context` automatically applied in this case, or not applicable?

rholinshead: Ah nvm, I see `lifespan_context` from request context used directly in those cases

saqadri: This is one of the hoops I had to jump through to get this to work. I think it should work as you pointed out using the lifespan_context

PR #255, opened 2025-06-06
Add support for sampling

StreetLamb:

PR #386, opened 2025-08-26
Sampling

roman-van-der-krogt:
Based on https://github.com/lastmile-ai/mcp-agent/pull/255.

- merged all latest changes
- made nested calls work
- add an example with documentation

## Summary by CodeRabbit

- New Features
  - Added haiku generation prompt/tool and a friends data resource.
  - Introduced human-in-the-loop sampling with local or upstream handling.
  - Added workflow tools (list, run, status, resume, cancel) and server creation utility.
  - Provided asyncio and Temporal demo servers/clients, plus SSE exposure.
  - Enabled context-aware server connections and automatic agent context initialization.

- Documentation
  - New sampling example README; updated resources README.
  - Added example configs and requirements.

- Tests
  - Removed outdated upstream session tests.

### DIFF
--- a/src/mcp_agent/mcp/mcp_server_registry.py
+++ b/src/mcp_agent/mcp/mcp_server_registry.py
@@ -33,7 +33,6 @@ from mcp_agent.config import (
 from mcp_agent.logging.logger import get_logger
 from mcp_agent.mcp.mcp_agent_client_session import MCPAgentClientSession
 from mcp_agent.mcp.mcp_connection_manager import MCPConnectionManager
-from mcp.server.session import ServerSession
 
 if TYPE_CHECKING:
     from mcp_agent.core.context import Context
@@ -158,7 +157,7 @@ class ServerRegistry:
                     read_stream,
                     write_stream,
                     read_timeout_seconds,
-                    context=context # XX WHICH CONTEXT?
+                    context=context
                 )
                 async with session:
                     logger.info(
@@ -296,7 +295,7 @@ class ServerRegistry:
             self,
             server_name: str,
             client_session_factory: Callable[
-                [MemoryObjectReceiveStream, MemoryObjectSendStream, timedelta, ServerSession],
+                [MemoryObjectReceiveStream, MemoryObjectSendStream, timedelta | None],
                 ClientSession,
             ] = ClientSession,
             init_hook: InitHookCallable = None,
--- a/src/mcp_agent/app.py
+++ b/src/mcp_agent/app.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, Optional, Type, TypeVar, Callable, TYPE_CHECKING
 from datetime import timedelta
 from contextlib import asynccontextmanager
 
+from mcp import ServerSession
 from mcp.server.fastmcp import FastMCP
 from mcp_agent.core.context import Context, initialize_context, cleanup_context
 from mcp_agent.config import Settings, get_settings
@@ -68,6 +69,7 @@ class MCPApp:
         human_input_callback: HumanInputCallback | None = None,
         elicitation_callback: ElicitationCallback | None = None,
         signal_notification: SignalWaitCallback | None = None,
+        upstream_session: Optional["ServerSession"] = None,
         model_selector: ModelSelector | None = None,
     ):
         """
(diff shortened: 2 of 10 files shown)


saqadri:
Looks like lint and test are failing -- can you please look into those?

examples/mcp/mcp_prompts_and_resources/README.md
```diff
@@ -21,9 +21,7 @@ This example demonstrates **both resources and prompts**.
 
   - **Resources:**
     - `demo://docs/readme`: A sample README file (Markdown)
-    - `demo://config/settings`: Example configuration settings (JSON)
-    - `demo://data/users`: Example user data (JSON)
-    - `demo://status/health`: Dynamic server health/status info (JSON)
```
saqadri, line 26: nit: any reason for the removal?

src/mcp_agent/mcp/mcp_server_registry.py
```diff
@@ -103,13 +105,14 @@ def load_registry_from_file(
...
-        session_id: str | None = None,
+            self,
+            server_name: str,
+            client_session_factory: Callable[
+                [MemoryObjectReceiveStream, MemoryObjectSendStream, timedelta | None, ServerSession | None],
+                ClientSession,
+            ] = ClientSession,
+            session_id: str | None = None,
+            upstream_session: ServerSession | None = None,
     ) -> AsyncGenerator[ClientSession, None]:
```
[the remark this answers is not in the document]
saqadri: This warning seems legit

src/mcp_agent/mcp/mcp_server_registry.py
```diff
@@ -134,6 +137,10 @@ async def start_server(
             else None
         )
 
+        # Record the upstream session (if any) in the context
+        from mcp_agent.core.context import Context
+        ctx_with_session = Context(upstream_session=upstream_session)
+
```
saqadri: I'm just thinking through the consequences of creating our own Context object here, vs. either making ServerRegistry a `ContextDependent`, or simply passing the `upstream_session` directly into the `client_session_factory`.

Any thoughts? The idea for Context is there is one Context object that is passed  through the lifetime of the MCPApp to anything that needs Context -- it doesn't get created by anyone other than `MCPApp` or if unspecified, loaded as a global by `ContextDependent`

src/mcp_agent/agents/agent.py
```diff
@@ -189,6 +189,20 @@ async def attach_llm(
...
+                from mcp_agent.core.context import get_current_context
+                self.context = get_current_context()
+
+            # Prefer a helper if/when Context switches to task-local binding
+            if hasattr(self.context, "set_active_llm"):
+                self.context.set_active_llm(self.llm)
+            else:
+                self.context.active_llm = self.llm
+
+            self.context.active_llm = self.llm
```
saqadri: Hmmm, this is a bit concerning because `Context` is global to the `MCPApp`, so any agent that is instantiated will set the "active_llm", which means the last agent that was instantiated is set as the "active_llm".

I actually think we can remove the concept of active_llm entirely and instead use the `ModelSelector` class to get the LLM that best matches the user's preferences.

Then when a sampling request comes in, sampling_handler can use `factory.py` to construct an AugmentedLLM with the right configuration to handle the sampling request.

If we want to limit sampling to JUST the agents that have been created, we can have context keep track of a Set/List of `available_llms` which can be used to filter to the closest one.

examples/mcp/mcp_prompts_and_resources/README.md
roman-van-der-krogt: The demo server doesn't have these resources. This makes the README consistent with the code

saqadri: Got it, thanks!

src/mcp_agent/mcp/mcp_server_registry.py
roman-van-der-krogt: Been pondering both options this morning, and I don't think making the ServerRegistry `ContextDependent` makes sense. Different workflows will have unique `upstream_session`s, so storing it in a global context doesn't work.

saqadri: > Different workflows will have unique upstream_sessions

Are you thinking of the case where the client reconnects? The `ServerRegistry` is 1:1 with the `MCPApp`, so in theory should always have a single upstream_session, but I haven't thought about it in too much detail, so would love to get your thoughts! I think keeping it to passing in the upstream_session is also pretty clean so based on your thinking here I'm good with either approach

saqadri:
Closing this PR as it's been resolved with other changes

PR #394, opened 2025-08-28
Add user parameter support to AzureAugmentedLLM

Tanmay-Kant:
Defined the user property in the RequestParams object which is passed as a request_params field on generate, generate_str, and generate_structured in order to facilitate communication to Azure OpenAI endpoints.


saqadri (approved):
Awesome, thank you for the contribution @Tanmay-Kant! This will be in a patch release tomorrow

PR #398, opened 2025-08-29
Use more flexible structured output mode for openai

StreetLamb:
In strict mode, OpenAI's structured output only supports a subset of JSON schema spec. Allow option to toggle between strict and non-strict mode via `RequestParams`.

## Summary by CodeRabbit

- New Features
  - Added an optional “strict” mode for structured completion requests to enforce exact schema conformity when needed.
- Behavior/Performance
  - Default behavior is now non-strict for structured completions, allowing more flexible responses unless strict mode is explicitly enabled.


saqadri:
Ship it!

Issue #393, opened 2025-08-27
Pass custom parameters to AzureAugmentedLLM request body


pavelorehov:
I need to pass extra custom parameters related to user key in request body to LLM.
Below is example with "user" parameter.
Cannot find any example if AzureAugmentedLLM supports that, if not would be good to add it or share workaround.

{
    "messages": [
        {
            "role": "system",
            "content": "You are a chatbot"
        },
        {
            "role": "user",
            "content": "who is the president of USA."
        }
    ],
    "user": "{\"appkey\": \"<appkey>\"}",
    "stop": [
        "<|im_end|>"
    ]
}

saqadri:
@pavelorehov thanks for reporting! This is supported by the OpenAIAugmentedLLM -- 

https://github.com/lastmile-ai/mcp-agent/blob/89b888eadb97caac1ebf7ca89d37fb260c555850/src/mcp_agent/workflows/llm/augmented_llm_openai.py#L220

But not the AzureAugmentedLLM doesn't support this. It should be the exact same pattern to add. Would you be open to contributing the fix? 

In order to specify it on the request, you just define the `user` property in the `RequestParams` object:

https://github.com/lastmile-ai/mcp-agent/blob/89b888eadb97caac1ebf7ca89d37fb260c555850/src/mcp_agent/workflows/llm/augmented_llm.py#L124

which is passed as a `request_params` field on generate, generate_str, and generate_structured:

https://github.com/lastmile-ai/mcp-agent/blob/89b888eadb97caac1ebf7ca89d37fb260c555850/src/mcp_agent/workflows/llm/augmented_llm.py#L175

Tanmay-Kant:
Hey @saqadri, I've been working with @pavelorehov on this, can I make the fix?

saqadri:
> Hey @saqadri, I've been working with @pavelorehov on this, can I make the fix?

Absolutely! @Tanmay-Kant just let me know how I can help

Tanmay-Kant:
Hey @saqadri, made the changes: https://github.com/lastmile-ai/mcp-agent/pull/394
I also added a unit test but did not push that in case we want to leave those alone. Please advise.


# Test 25: User Parameter Handling
    @pytest.mark.asyncio
    async def test_user_param_handling(self, mock_llm, default_usage):
        """
        Tests that the user parameter is properly included in Azure API calls.
        """
        # Setup mock executor
        mock_llm.executor.execute = AsyncMock(
[...]

PR #403, opened 2025-09-02
Fix Azure user parameter handling in request_completion_task

Tanmay-Kant:
## Summary

This PR fixes an issue where the user parameter was not being properly transmitted to Azure OpenAI API calls.

## Changes

- Modified `AzureCompletionTasks.request_completion_task()` to properly extract the user parameter from the payload
- Pass user parameter as a separate kwarg to `azure_client.complete()` instead of including it in the payload
- Ensures user parameter is handled according to Azure SDK expectations

## Technical Details

The Azure SDK expects the `user` parameter to be passed as a keyword argument to the `complete()` method, not as part of the request payload. This change:

1. Extracts the user parameter from the payload using `payload.pop('user', None)`
2. Passes it separately as `user=user` kwarg when calling `azure_client.complete()`

## Related

- Related to issue #393 where custom parameters need proper handling in Azure requests
- Builds upon the user parameter support added in previous PR #394

## Testing

This fix ensures that user attribution works correctly for Azure OpenAI API calls when the user parameter is provided via RequestParams or Azure config.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures user context is sent correctly to Azure, improving accuracy and auditability.
  * Prevents unintended payload mutations that could affect subsequent requests.
* **Performance**
  * Improves responsiveness by executing Azure requests in a background thread, reducing blocking during LLM operations.

### DIFF
--- a/src/mcp_agent/workflows/llm/augmented_llm_azure.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_azure.py
@@ -539,12 +539,20 @@ class AzureCompletionTasks:
                 ),
             )
 
-        payload = request.payload
+        payload = request.payload.copy()
+        # Remove user from kwargs - Azure SDK expects it in request body, not as kwarg
+        user = payload.pop('user', None)
+        
         # Offload sync SDK call to a thread to avoid blocking the event loop
         loop = asyncio.get_running_loop()
-        response = await loop.run_in_executor(
-            None, functools.partial(azure_client.complete, **payload)
-        )
+        if user:
+            response = await loop.run_in_executor(
+                None, functools.partial(azure_client.complete, user=user, **payload)
+            )
+        else:
+            response = await loop.run_in_executor(
+                None, functools.partial(azure_client.complete, **payload)
+            )
         return response
 
 


src/mcp_agent/workflows/llm/augmented_llm_azure.py
```diff
@@ -539,12 +539,20 @@ async def request_completion_task(
                 ),
             )
 
-        payload = request.payload
+        payload = request.payload.copy()
```
saqadri, line 542: Why do we need to copy the payload object? Should be fine to pop it directly since this function is the only one using it. 

Since the payload can contain large amounts of messages it would be ideal not to have tocopy it on every request

saqadri (changes requested):
Generally looks good, except for one issue. Putting it back on your queue and then we can get it merged.

saqadri:
@Tanmay-Kant can you please let me know if you're able to resolve the PR feedback, or if you'd like help taking this to completion. Thanks again for the contribution! Keen to get your contribution in, so just let me know.

PR #414, opened 2025-09-04
Temporal MCP upstream-session proxy

roman-van-der-krogt:
Temporal workers run outside the original MCP request context; they need a reliable way to:
  - stream logs back to the MCP client
  - elicit human input,
  - request sampling from MCP client
  - and send other notifications to the MCP client.
- Determinism constraints mean workflows can’t perform non-deterministic I/O; activities must “phone home” to the server gateway.

### Overview of changes
Core idea is the introduction of a SessionProxy, which derives from `mcp.server.session.ServerSession` and for all intents and purposes appears as a server session inside a Temporal context, but in reality proxies requests through the MCPApp that *actually* has the upstream session to the MCP client.

There are a lot of moving parts to get this to work, including:

- Execution ID and routing
  - `get_execution_id()` now derives from Temporal runtime:
    - In workflow: `workflow.info().run_id`
    - In activity: `activity.info().workflow_run_id`
    - Else: fallback to a module-global (non-Temporal contexts only)

- ContextInterceptor
  - Client/workflow outbound sets `__execution_id` header from current context.
  - Workflow/activity inbound restores the global execution_id for any code running off the Temporal trampoline.

- Session proxying
  - `SessionProxy` subclasses `mcp.server.session.ServerSession`.
  - Inside workflows: forwards logs via the `"mcp_forward_log"` activity (deterministic).
  - Outside workflows: uses the HTTP gateway helpers.
  - Minimal `rpc` facade provided for `request/notify` parity.

- Server gateway endpoints on App server
  - Private routes: `/internal/workflows/log`, `/internal/session/by-run/{run_id}/notify`, `/internal/session/by-run/{run_id}/request`, `/internal/human/prompts`.
  - Optional `X-MCP-Gateway-Token` authentication.
  - Idempotency support for notify (idempotency_key) to handle retries.
  - Cleaned up noisy debug logging; behavior unchanged.

- Logging improvements
  - Default-bound app context for loggers created after app init (to pick up `upstream_session` automatically).
  - Removed unsafe `get_current_context` fallback.
  - Workflow runtime path schedules the log-forward activity when no upstream_session is available.
  - Upstream listener no longer prints “skip” messages in normal non-worker paths.

- Tests
  - Added: execution_id resolution (workflow/activity) and interceptor restoration.
  - Added: proxy HTTP helper tests for ok/error paths and token handling.
  - Removed two brittle unit tests that exercised unexported server wiring and a workflow-only logging branch without a Temporal runtime (these caused false negatives); the remaining temporal tests still pass.

#### Security
- Optional gateway token enforcement via `X-MCP-Gateway-Token`.
- Notify idempotency prevents duplicated side effects on retries.

### To Test
- Run the temporal example client:
  - `uv run examples/mcp_agent_server/temporal/client.py --server-log-level debug`
  - Observe “[SERVER LOG] …” notifications flowing during long-running workflows.

## Summary by CodeRabbit

- New Features
  - Server log streaming to clients with configurable server log level and end-to-end server->client log propagation.
  - Per-run gateway overrides (URL/token) via workflow memo; external gateway proxy endpoints for forwarding logs/requests/prompts.
  - Human input tool to submit replies to running workflows; workflows can forward logs, notifications, and requests via a session proxy.
  - Automatic execution-id propagation across workflows, activities, and client sessions.

- Improvements
  - Safer non-blocking logging in workflow runs with stderr fallback and default-bound context; improved log-tail error handling in CLI.

- Documentation
  - Updated README diagram and expanded CLI docstrings.

- Tests
  - New tests covering execution-id propagation, HTTP proxy helpers, and per-server tool filtering.

- Chores
  - Version bumped to 0.1.17.


src/mcp_agent/executor/workflow.py
```diff
@@ -16,8 +16,14 @@
 
 from pydantic import BaseModel, ConfigDict, Field
 from mcp_agent.core.context_dependent import ContextDependent
-from mcp_agent.executor.workflow_signal import Signal, SignalMailbox
+from mcp_agent.executor.temporal import TemporalExecutor
```
saqadri: This one is probably a merge conflict resolution issue... this should be only in TYPE_CHECKING so that `workflow.py` doesn't have a direct Temporal dependency. Otherwise even if a user just wants asyncio they still need to import temporal package.

src/mcp_agent/mcp/client_proxy.py
```diff
@@ -0,0 +1,153 @@
...
+        return env_url.rstrip("/")
+
+    # Next: a registry entry (if provided)
+    if server_registry and server_name:
+        cfg = server_registry.get_server_context(server_name)
+        if cfg and getattr(cfg, "url", None):
+            return cfg.url.rstrip("/")
+
+    # Fallback: default local server
+    return "http://127.0.0.1:8000"
```
saqadri, line 31: Should we be doing this?

src/mcp_agent/executor/temporal/interceptor.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+class ContextPropagationInterceptor(
+    temporalio.client.Interceptor, temporalio.worker.Interceptor
+):
+    """Interceptor that propagates a value through client, workflow and activity calls.
+
+    This interceptor implements methods `temporalio.client.Interceptor` and  `temporalio.worker.Interceptor` so that
+
+    (1) a user ID key is taken from context by the client code and sent in a header field with outbound requests
```
rholinshead: Execution ID instead of user ID, right?

src/mcp_agent/core/context.py
```diff
@@ -89,6 +89,10 @@ class Context(BaseModel):
     # Token counting and cost tracking
     token_counter: Optional[TokenCounter] = None
 
+    # Dynamic gateway configuration (per-run overrides via Temporal memo)
```
rholinshead, line 92: I think this might make more sense to be done in TemporalSettings instead of the root context

src/mcp_agent/executor/workflow.py
```diff
@@ -734,6 +814,64 @@ async def initialize(self):
...
+
+                if isinstance(memo_map, dict):
+                    gw = memo_map.get("gateway_url")
+                    gt = memo_map.get("gateway_token")
+
+                    self._logger.debug(
+                        f"Proxy parameters: gateway_url={gw}, gateway_token={gt}"
+                    )
+
+                    if gw:
```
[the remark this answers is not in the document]
rholinshead: +1

src/mcp_agent/logging/logger.py
```diff
@@ -15,6 +15,8 @@
 
 from contextlib import asynccontextmanager, contextmanager
 
+import temporalio
+
```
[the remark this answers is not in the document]
rholinshead: This will be similar to above comment, we don't want to make temporalio a hard dependency for mcp-agent, only when the execution engine is temporal

src/mcp_agent/core/context.py
saqadri: @rholinshead I think you're thinking of the config, not context. This does need to be a runtime value, though possibly the value could be loaded from a config in the future. IMO this is fine as is.

PR #415, opened 2025-09-04
feat: multiple otel exporters

saqadri:
Clean up `OpenTelemetrySettings` to be able to specify multiple exporters. This allows you to output to console, file, and even multiple otlp collector endpoints. 

Particularly nice for MCP-agent cloud, where a developer can configure their agent to export to their own OTEL endpoint (while still allowing us to configure out internal OTLP endpoint for surfacing logs to them).

## Summary by CodeRabbit

- **New Features**
  - Added configurable tracing sampling rate.
  - Introduced typed OpenTelemetry exporters (console, file, OTLP) with per-exporter settings and support for multiple exporters.
  - Enhanced file exporter with dynamic path patterns and new log path settings.

- **Refactor**
  - Modernized configuration schema to use typed exporters; deprecated legacy otlp_settings/path fields (backward-compatible coercion).
  - Removed allowed_tools from MCP server configuration.

- **Documentation**
  - Updated tracing examples and READMEs with typed exporter workflow, OTLP endpoint examples, and expanded setup steps.


schema/mcp-agent.config.schema.json
```diff
@@ -912,30 +996,7 @@
...
-          "anyOf": [
-            {
-              "$ref": "#/$defs/TracePathSettings"
-            },
-            {
-              "type": "null"
-            }
-          ],
-          "default": null
+          "description": "Deprecated single OTLP settings. Prefer exporters list with type \"otlp\"."
```
rholinshead, line 1051: What's the intention for supporting the deprecated settings vs just enforcing the new ones going forward?

src/mcp_agent/config.py
```diff
@@ -105,10 +112,6 @@ class MCPServerSettings(BaseModel):
     env: Dict[str, str] | None = None
     """Environment variables to pass to the server process."""
 
-    allowed_tools: Set[str] | None = None
```
rholinshead, line 109: Intentional?

src/mcp_agent/config.py
```diff
@@ -462,15 +465,60 @@ class TraceOTLPSettings(BaseModel):
...
+
+class FileExporterSettings(OpenTelemetryExporterBase):
+    type: Literal["file"] = "file"
+    path: str | None = None
+    path_settings: TracePathSettings | None = None
+
+
+class OTLPExporterSettings(OpenTelemetryExporterBase):
+    type: Literal["otlp"] = "otlp"
+    endpoint: str | None = None
```
rholinshead, line 512: Should this be required str?

examples/tracing/langfuse/mcp_agent.secrets.yaml.example
```diff
@@ -7,6 +7,7 @@ anthropic:
   api_key: anthropic_api_key
 
 otel:
+  # Headers are merged with typed OTLP exporter settings
```
rholinshead, line 10: How would this look for the new format? Would we need to duplicate the `-type` and `endpoint` here to map to the correct one (in case of multiple)?

src/mcp_agent/mcp/mcp_aggregator.py
```diff
@@ -345,26 +345,37 @@ async def load_server(self, server_name: str):
...
+                    or not hasattr(self.context.server_registry, "get_server_config")
+                ):
+                    logger.warning(
+                        f"No config found for server '{server_name}', no tool filter will be applied..."
+                    )
                 else:
-                    allowed_tools = self.context.server_registry.get_server_config(server_name).allowed_tools
+                    allowed_tools = self.context.server_registry.get_server_config(
+                        server_name
+                    ).allowed_tools
```
rholinshead, line 354: Isn't this removed above now?

schema/mcp-agent.config.schema.json
saqadri: Deprecating for one release to give people time to migrate, and then will remove in a subsequent change

saqadri: Just best practice for a package. Can remove now as well, but it's relatively lightweight to support back-compat right now

src/mcp_agent/config.py
saqadri: oof good catch thanks!

src/mcp_agent/config.py
saqadri: Good catch, yes

examples/tracing/langfuse/mcp_agent.secrets.yaml.example
saqadri: Yes, `headers` would need to be specified for each endpoint, which I think is correct.

```
class OTLPExporterSettings(OpenTelemetryExporterBase):
    type: Literal["otlp"] = "otlp"
    endpoint: str | None = None
    headers: Dict[str, str] | None = None
```

PR #435, opened 2025-09-08
Feature: Update Workflow Tool Calls to Work with workflow_id

rholinshead:
## Summary by CodeRabbit

- New Features
  - Workflow actions now accept either run_id or workflow_id; when only workflow_id is provided, operations target the latest run.
  - Unified status endpoint supports lookups by run_id or workflow_id.
  - In-memory registry can list all workflows.

- Refactor
  - Simplified public tools: removed per-workflow get_status; use the unified status tool instead.
  - IDs are used consistently across run, status, resume, and cancel flows.

- Documentation
  - Updated help and docs to describe optional IDs and “latest run” behavior.

- Tests
  - Expanded coverage for ID-based lookups and latest-run semantics.

- Chores
  - Example configs use environment-resolved binaries.

## Testing
```bash
make lint
make format
make tests
```

- Ran all examples in `examples/mcp_agent_server/asyncio` and `examples/mcp_agent_server/temporal`, including running and connecting to sse server in mcp inspector to manually call workflow tools with run_id and with workflow_id, verifying they work with each ID
- Ran all examples in `examples/temporal`, a couple needing fixing. Note router.py example is failing with `NotFoundError` on main and here

### DIFF
--- a/src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
+++ b/src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
 logger = get_logger(__name__)
 
 
-class QualityRating(str, Enum):
+class QualityRating(int, Enum):
     """Enum for evaluation quality ratings"""
 
     POOR = 0  # Major improvements needed


examples/mcp_agent_server/asyncio/mcp_agent.config.yaml
```diff
@@ -7,16 +7,12 @@ logger:
 mcp:
   servers:
     fetch:
-      command: "/Users/saqadri/.local/bin/uvx"
```
saqadri, line 10: thank you!

src/mcp_agent/executor/temporal/workflow_registry.py
```diff
@@ -61,30 +61,52 @@ async def unregister(self, run_id: str, workflow_id: str | None = None) -> None:
...
         await self._executor.ensure_client()
 
         try:
-            workflow = await self.get_workflow(run_id)
-            workflow_id = (
-                workflow.name if workflow and workflow_id is None else workflow_id
-            )
+            workflow = await self.get_workflow(run_id, workflow_id)
+            if workflow and not workflow_id:
+                workflow_id = workflow.id or workflow.name
```
saqadri, line 94: the previous behavior was effectively `workflow.name or workflow.id`... IIRC that was intentional but I don't remember why anymore. LMK if you noticed any issues

src/mcp_agent/server/app_server.py
```diff
@@ -668,6 +668,8 @@ def list_workflows(ctx: MCPContext) -> Dict[str, Dict[str, Any]]:
                 endpoints = [
                     f"workflows-{workflow_name}-run",
                     f"workflows-{workflow_name}-get_status",
+                    f"workflows-{workflow_name}-cancel",
+                    f"workflows-{workflow_name}-resume",
```
saqadri: Let's remove these

src/mcp_agent/server/app_server.py
```diff
@@ -1426,20 +1445,27 @@ async def _workflow_run(
...
     if not workflow_registry:
         raise ToolError("Workflow registry not found for MCPApp Server.")
 
-    workflow = await workflow_registry.get_workflow(run_id)
-    workflow_id = workflow.id if workflow and workflow.id else workflow_name
+    if not workflow_id:
+        workflow = await workflow_registry.get_workflow(
+            run_id=run_id, workflow_id=workflow_id
+        )
+        workflow_id = workflow.id if workflow else None
```
saqadri: This is also changing the behavior. Should it still default to workflow_name? 
```suggestion
        workflow_id = workflow.id if workflow and workflow.id else workflow_name
```

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
```diff
@@ -22,7 +22,7 @@
 logger = get_logger(__name__)
 
 
-class QualityRating(str, Enum):
+class QualityRating(int, Enum):
```
saqadri, line 25: what's the reason for this change?

tests/executor/temporal/test_workflow_registry.py
```diff
@@ -112,3 +112,164 @@ async def test_list_workflows(registry):
...
+        ValueError, match="Either run_id or workflow_id must be provided"
+    ):
+        await registry.cancel_workflow()
+
+
+@pytest.mark.asyncio
+async def test_get_workflow_status_by_workflow_id(registry, mock_executor):
+    mock_workflow = MagicMock(name="test_workflow")
+    mock_workflow.id = "workflow-id"
+    mock_workflow.name = "workflow-id"
```
saqadri: The tests should actually have a different value for workflow.name and .id, because I am worried about inconsistent use. This is a tricky part of the codebase. Workflow Name would be the class name or tool name, and workflow id would be a uuid

saqadri (approved):
Approving to unblock, but worried about the change in behavior where we use the name vs. id.

src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
rholinshead: On main evaluator_optimizer is broken because the LLM is returning an str (e.g. "GOOD") and then, I think, the updated openai structured serialization is erroring because we expect a number for the enum value and get back a string. Changing this to int results in proper serialization

saqadri: ah I see

rholinshead: I can undo and leave as failing for now, just wanted to make sure the example worked

saqadri: No let's go with it, that's totally fine

src/mcp_agent/executor/temporal/workflow_registry.py
rholinshead: I haven't noticed any issues. Looking at the register/unregister it's already inconsistent in how workflow_id and name are used. Registration happens with workflow_id (which is optional, so potential for None errors). But then unregister will prioritize name if no workflow_id is given... My understanding is it's intended to just be a reference to map all the run_ids for the workflow so realistically using one or the other won't matter as long as it's consistent? The default asyncio workflow id is created as `self._workflow_id = provided_workflow_id or self.name`, so it uses what's provided at run call or defaults to the name. So id and name are synonymous for asyncio case unless workflow_id is provided, in which case we _do_ want to prioritize id over name

rholinshead: I can fix the register issue / consistency here

PR #442, opened 2025-09-09
fix: logger cleanup and fallback kill

alienzach:


alienzach:
* **#442** <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/442?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -504,13 +504,17 @@ async def generate_structured(
             if params.metadata:
                 payload.update(params.metadata)
 
-            completion: ChatCompletion = await self.executor.execute(
```
saqadri, line 507: why remove the type hint?

src/mcp_agent/workflows/llm/augmented_llm_openai.py
```diff
@@ -504,13 +504,17 @@ async def generate_structured(
...
+            completion = await self.executor.execute(
                 OpenAICompletionTasks.request_completion_task,
                 RequestCompletionRequest(
                     config=self.context.config.openai, payload=payload
                 ),
             )
 
+            # Check if the executor returned an exception
+            if isinstance(completion, Exception):
```
saqadri, line 515: Does this happen?

alienzach: during testing, when completion api call throws exception, `completion` variable will be an exception instead of data object. this will result on L518's reference to `choices` to throw `AtributeErr` due to the `choices` field missing from the exception. 

this change is to catch the exception and thrown it for clear error messaging.

src/mcp_agent/workflows/llm/augmented_llm_openai.py
alienzach: this `completion` might either be data object in normal flow or exception if API error

saqadri: Then we should change the type to `ChatCompletion | Exception`, I don't want to remove type hints since they are really useful for autocomplete

src/mcp_agent/app.py
```diff
@@ -391,6 +402,75 @@ async def cleanup(self):
         self._initialized = False
         self._tracer_provider = None
 
+    async def _force_cancel_remaining_tasks(self) -> None:
```
saqadri, line 405: I'm not sure about the safety of this function. Specifically filtering this way gives me pause because I'm not sure of the implications of this:

```
# Consider tasks under our package path only
                    return "/mcp_agent/" in filename.replace("\\", "/")
```

PR #447, opened 2025-09-11
pass api key to workflow

roman-van-der-krogt:
Pass the API token to the upstream temporal worker for calling the `internal` APIs with.

## Summary by CodeRabbit

- Bug Fixes
  - Workflow runs now fall back to the configured Temporal API key when no gateway token is provided, ensuring tokens are included in workflow memos when available; header handling and error behavior remain unchanged. No public API changes.

- Style
  - Minor formatting cleanup with no behavioral impact.

### DIFF
--- a/src/mcp_agent/server/app_server.py
+++ b/src/mcp_agent/server/app_server.py
@@ -1399,7 +1399,10 @@ async def _workflow_run(
         try:
             # Prefer explicit kwargs, else infer from request context/headers
             gateway_url = kwargs.get("gateway_url")
-            gateway_token = kwargs.get("gateway_token") or app.config.temporal.api_key
+            gateway_token = kwargs.get("gateway_token")
+            if gateway_token is None:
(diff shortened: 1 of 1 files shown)


src/mcp_agent/server/app_server.py
```diff
@@ -1399,7 +1399,7 @@ async def _workflow_run(
         try:
             # Prefer explicit kwargs, else infer from request context/headers
             gateway_url = kwargs.get("gateway_url")
-            gateway_token = kwargs.get("gateway_token")
+            gateway_token = kwargs.get("gateway_token") or app.config.temporal.api_key
```
saqadri: Would just null-check app.config.temporal, but otherwise looks good!

roman-van-der-krogt: yep, that caused the test cases to fail, in fact

saqadri: nice those are some good tests then

PR #467, opened 2025-09-12
Fix: Update Deployment Secrets Processing UX

rholinshead:
## Description
Update the deployment secrets processing UX to respect `mcp_agent.secrets.yaml` as a "readonly" artifact, dropping the manual custom secrets tag / environment variable handling in favour of raw secrets transformation into the `mcp_agent.deployed.secrets.yaml`. This also replaces the 'developer secret' naming with 'deployment secret' to be clearer in its use.

More specifically, this means:
- users no longer need to manually modify their secrets file to add `!developer_secret` or `!user_secret` tags, or specify env var names for such secrets; they use their secrets file as prescribed by `mcp-agent` without special handling for cloud
- during deployment, we first check if `mcp_agent.deployed.secrets.yaml` exists; if so, prompt if it should be reprocessed (if not, it will be used)
- during processing, prompt the user to update any deployed secrets that don't match (if both files exist)
- during processing, prompt the user if a secret is a deployment secret or user secret, e.g.
```bash
Select secret type for 'openai.api_key'
1: Deployment Secret: The secret value will be stored securely and accessible to the deployed application runtime.
2: User Secret: No secret value will be stored. The 'configure' command must be used to create a configured 
application with this secret.

Select secret type: (1): 
```
- the user can skip prompting via the `--non-interactive` flag and the command will either just use the existing deployed secrets file as-is (if it exists) or create new deployment secrets for all secrets in the `mcp_agent.deployed.secrets.yaml` file

Removed some redundant tests and added some new ones for the transformation logic update. Removed --dry-run flag for now since it was broken. We should consider adding a new --dry-run implementation which spins up a local mcp app server and does a test list_tools call over sse.

## Testing
```bash
make lint
make format
make tests
```

Ran through the various flows and ensured secrets handling works as expected

## Summary by CodeRabbit

* New Features
  * Interactive secret classification (Deployment vs User) with reuse of existing deployed secrets.
  * Non-interactive mode defaults all secrets to deployment; clearer console summaries (created, reused, skipped).
  * Prompts when updating existing apps; prints API target info; deployment header shows App, ID, and deployed secrets file.

* Refactor
  * Simplified deploy flow: removed --dry-run and --no-secrets; API URL and key now required.
  * Shift from developer to deployment secrets; stricter error handling.
  * Config discovery now includes deployed secrets file to drive readiness and processing.


saqadri:
Going to merge this in so I can test it!

PR #471, opened 2025-09-14
Query temporal directly for workflow runs

saqadri:
Currently `workflow-runs-list` tool only works for the in-memory state of the MCPApp, which won't surface previous runs or persist across restarts. This change makes it work by querying temporal directly.

## Summary by CodeRabbit

- New Features
  - Workflow runs listing now supports pagination (limit, page size, next page token) across server and client.
  - Workflow status entries now include workflow_id and run_id.
  - CLI supports authentication via MCP_API_KEY in addition to stored credentials; credentials also discovered in XDG config paths.

- Bug Fixes
  - More robust credential handling: atomic saves, clearer warnings on corrupted files, and cleanup across all locations.

- Refactor
  - CLI app commands moved under “apps” namespace (mcp-agent apps …); legacy apps group removed.
  - Cloud workflow commands use async server resolution with improved errors for missing server URLs.


src/mcp_agent/executor/temporal/workflow_registry.py
```diff
@@ -216,23 +216,140 @@ async def get_workflow_status(
...
-        return result
+        # Collect all executions for this task queue (best effort)
+        try:
+            await self._executor.ensure_client()
+            client = self._executor.client
+
+            # Use caller query if provided; else default to newest first
+            query_local = query or "order by StartTime desc"
+
+            iterator = client.list_workflows(
```
rholinshead, line 260: Just noting for later -- when we push support for multiple users using a server, we'll need to add the auth checks in the proxy layer for this request, which will have implications for results returned (e.g. we'll only return those that the user has permission to see)

saqadri: Good point, let me mark that as a TODO comment in the code as well. I think this is reasonable for the time being but agreed we'll want this to be user-scoped in the future.

src/mcp_agent/executor/temporal/workflow_registry.py
```diff
@@ -216,23 +216,140 @@ async def get_workflow_status(
...
+                        temporal_status.get("status")
+                        if isinstance(temporal_status, dict)
+                        else None
+                    )
+                    if isinstance(ts, str):
+                        status_dict["status"] = ts.lower()
+                        status_dict["running"] = ts.upper() in {"RUNNING", "OPEN"}
+                except Exception:
+                    pass
+
```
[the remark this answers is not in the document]
rholinshead: Could be legit if there is a large count of workflows; list workflows seems to already get most of the describe info from the resulting WorkflowExecution entity, right?

src/mcp_agent/executor/workflow_registry.py
```diff
@@ -267,12 +298,33 @@ async def get_workflow_status(
...
+        workflows = list(self._workflows.values()) if self._workflows else []
+        try:
+            workflows.sort(
+                key=lambda wf: (wf.state.updated_at if wf.state else None) or 0,
+                reverse=True,
+            )
+        except Exception:
+            pass
+
+        result: List[Dict[str, Any]] = []
```
rholinshead, line 321: Ok to skip pagination for default registry?

src/mcp_agent/executor/temporal/workflow_registry.py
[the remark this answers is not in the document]
saqadri: Will investigate

src/mcp_agent/executor/workflow.py
```diff
@@ -586,6 +586,8 @@ async def get_status(self) -> Dict[str, Any]:
         """
         status = {
             "id": self._run_id,
+            "workflow_id": self.id,
```
rholinshead, line 589: As part of testing can you please double-check that serialization works for the workflow commands, e.g. `uv run mcp-agent workflows runs <server_url>`

src/mcp_agent/executor/temporal/workflow_registry.py
rholinshead: Fwiw, this client code shouldn't really need to know/care about it, as long as our rpc metadata passes through the api key. Our backend proxy layer will handle the auth checks, but will need to handle pagination and such with some in-memory processing

src/mcp_agent/executor/workflow_registry.py
saqadri: Yes that should be fine IMO. At least for now

src/mcp_agent/executor/workflow.py
saqadri: Tested this and works, but also made some updates so that we handle the case where we return a next_page_token (though we don't actually use that token in the CLI yet)

src/mcp_agent/executor/temporal/workflow_registry.py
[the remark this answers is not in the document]
saqadri: Thanks, fixed

PR #473, opened 2025-09-14
Making MCP Server Aggregator Cloud ready

andrew-lastmile:
Local run:
image

Cloud run:
image

To do: 
- investigate why this example fails to start file-system server

### DIFF
--- a/examples/basic/mcp_server_aggregator/main.py
+++ b/examples/basic/mcp_server_aggregator/main.py
@@ -8,8 +8,16 @@ from rich import print
 
 app = MCPApp(name="mcp_server_aggregator")
 
-
-async def example_usage_persistent():
+@app.tool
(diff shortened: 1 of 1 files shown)


andrew-lastmile:
non-persistent mode fails for the filesystem mcp server:

```
Filesystem server (@modelcontextprotocol/server-filesystem):
	•	Stateful by design.
	•	Needs to negotiate roots (allowed directories) with the client at initialization.
	•	Exposes a larger toolset (read_text_file, list_directory, etc.), which depends on knowing its roots.
	•	In non-persistent mode, the aggregator spins it up, but there’s no stable channel to handle that roots handshake — so it errors out:
	•	Failed to validate request: 'NoneType' object has no attribute 'roots'
	•	Tool 'read_text_file' not found
```
	
non-blocking issue, merging for now

PR #501, opened 2025-09-17
feat: optional set global setting

alienzach:


alienzach:
* **#501** <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/501?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/501?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/mcp_agent/config.py
```diff
@@ -724,8 +724,18 @@ def _clear_global_settings():
...
-def get_settings(config_path: str | None = None) -> Settings:
-    """Get settings instance, automatically loading from config file if available."""
+def get_settings(config_path: str | None = None, set_global: bool = True) -> Settings:
+    """Get settings instance, automatically loading from config file if available.
+
+    Args:
+        config_path: Optional path to config file. If None, searches for config automatically.
+        set_global: Whether to set the loaded settings as the global singleton.
+                   Default is True for backward compatibility. Set to False for
+                   multi-threaded environments to avoid global state modification.
```
saqadri: Formatting looks a bit off.

```suggestion
          Default is True for backward compatibility. Set to False for
          multi-threaded environments to avoid global state modification.
```

src/mcp_agent/config.py
```diff
@@ -742,7 +752,8 @@ def deep_merge(base: dict, update: dict) -> dict:
         return merged
 
     global _settings
-    if _settings:
+    # Only return cached global settings if we're in set_global mode and no specific path requested
+    if _settings and set_global and config_path is None:
```
saqadri: Why the `config_path` check? That looks incorrect

```suggestion
    if _settings and set_global:
```

src/mcp_agent/config.py
```diff
@@ -742,7 +752,8 @@ def deep_merge(base: dict, update: dict) -> dict:
         return merged
 
     global _settings
```
saqadri: This should probably be inside the if block

tests/test_app.py
```diff
@@ -121,11 +121,16 @@ async def test_initialization_with_custom_settings(self):
         assert app._config is mock_settings
 
     @pytest.mark.asyncio
-    async def test_initialization_with_settings_path(self):
+    async def test_initialization_with_settings_path(self, tmp_path):
         """Test initialization with settings path."""
-        app = MCPApp(name="test_app", settings="path/to/settings.yaml")
+        # Create a temporary config file
+        config_file = tmp_path / "settings.yaml"
+        config_file.write_text("execution_engine: asyncio\n")
```
saqadri: Why are we doing this? Can we mock it instead?

src/mcp_agent/config.py
```diff
@@ -760,7 +771,10 @@ def deep_merge(base: dict, update: dict) -> dict:
             yaml_settings = yaml.safe_load(buf) or {}
 
             # Preload is authoritative: construct from YAML directly (no env overlay)
-            return Settings(**yaml_settings)
+            settings = Settings(**yaml_settings)
+            if set_global:
+                _settings = settings
```
rholinshead: Interesting, I suppose this was just missed before? I can't think of a reason why preload settings wouldn't be cached similarly to the others

src/mcp_agent/config.py
```diff
@@ -806,21 +820,26 @@ def deep_merge(base: dict, update: dict) -> dict:
                     yaml_secrets = yaml.safe_load(f) or {}
                     merged_settings = deep_merge(merged_settings, yaml_secrets)
 
-        _settings = Settings(**merged_settings)
-        return _settings
+        settings = Settings(**merged_settings)
+        if set_global:
+            _settings = settings
+        return settings
```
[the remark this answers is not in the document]
rholinshead: I suppose this PR is consistent with current logic of only warning in one case but I assume the others were just missed for the warning? Maybe we could do a helper,

```py
def _set_and_warn_global_settings(set_global: bool, settings: Settings):
    if set_global:
          global _settings
          if settings:
              _settings = settings
              # Thread-safety advisory: warn when setting global singleton from non-main thread
              if threading.current_thread() is not threading.main_thread():
                  warnings.warn(
                     "get_settings() is setting the global Settings singleton from a non-main thread. "
                    "In multithreaded environments, use get_settings(set_global=False) to avoid "
                    "global state modification, or pass the Settings instance explicitly to MCPApp(settings=...).",
                    stacklevel=2,
                )     
```

src/mcp_agent/config.py
alienzach: Make sense, we should change this to respect the original flow and return the object without setting global? If you don't mind could you share a pointer to the preload method?

rholinshead: Oh, I meant that this change is probably fine and was likely just missed earlier. The preload settings are configured via `MCP_APP_SETTINGS_PRELOAD` env var when running the app (see https://github.com/lastmile-ai/mcp-agent/pull/299)

PR #502, opened 2025-09-17
Sampling/Elicitation for workflows via sampling

roman-van-der-krogt:
- added a signal handler `_user_response` to (auto) workflows
- in `session_proxy`, if we're running in a temporal workflow, opt for the async request, and wait for the signal to arrive
- in `client_proxy`, call either the new async request (if requested), or the existing synchronous request
- in `app_server`, add a new asynchronous endpoint for sampling and elicitation. When called returns immediately after starting a task that executes on the request, then signals the result to the workflow
  - some major refactoring here, to extract common code between the two request methods
  - also refactored the auth requests, which were duplicated in each `internal` endpoint

Other things:
- new example for sampling/elicitation using (local) temporal
- in `sampling_handler`, add more details on the sampling request/response for the user to make an informed decision

## Summary by CodeRabbit

- New Features
  - Added an elicitation demo (tool, workflow, client, and worker) demonstrating booking + haiku flow with workflow status polling and a top-level runnable example.
  - Added async-request support for workflows to enable non-blocking RPC-style interactions.
  - LLM creation now accepts an agent name string.

- Bug Fixes / Improvements
  - Human approval prompts now include concise request/response summaries and metadata for clearer review.

- Refactor
  - Centralized gateway authentication and modularized internal request dispatch.

- Documentation
  - New example config and secrets templates for Temporal setups.

- Chores
  - Example deps updated to include Temporal SDK.


src/mcp_agent/executor/temporal/session_proxy.py
```diff
@@ -116,15 +119,27 @@ async def request(
             return {"error": "missing_execution_id"}
 
         if _in_workflow_runtime():
+            from temporalio import workflow
```
saqadri: We already import workflow as `_twf` at the start of the file

src/mcp_agent/executor/temporal/session_proxy.py
```diff
@@ -116,15 +119,27 @@ async def request(
...
             act = self._context.task_registry.get_activity("mcp_relay_request")
-            return await self._executor.execute(
+
+            await self._executor.execute(
                 act,
+                True,
```
saqadri: What does this True refer to? (worth adding a brief inline comment)

src/mcp_agent/executor/temporal/session_proxy.py
```diff
@@ -116,15 +119,27 @@ async def request(
...
+            await workflow.wait_condition(
+                lambda: _workflow_states.get(exec_id, {}).get('response_received', False)
+            )
+
+            return _workflow_states.get(exec_id, {}).get('response_data', {"error": "no_response"})
+
+        # Non-workflow (activity/asyncio): direct call and wait for result
         return await self._system_activities.relay_request(
-            exec_id, method, params or {}
+            False, exec_id, method, params or {}
```
saqadri: nit: 
```suggestion
            make_async_call=False, exec_id, method, params or {}
```

src/mcp_agent/executor/workflow.py
```diff
@@ -429,6 +429,27 @@ async def cancel(self) -> bool:
         from temporalio.common import RawValue
         from typing import Sequence
 
+        @workflow.signal()
+        async def _user_response(self, response: Dict[str, Any]):
+            """Signal handler that receives user responses."""
+            # Import here to avoid circular dependencies
+            try:
+                from temporalio import workflow
```
saqadri: This was imported on line 428

src/mcp_agent/executor/temporal/session_proxy.py
```diff
@@ -20,6 +20,9 @@
 from mcp_agent.executor.temporal.temporal_context import get_execution_id
 
 
+# Global state for signal handling (will be stored per workflow instance)
+_workflow_states: Dict[str, Dict[str, Any]] = {}
```
saqadri: Does this need to be protected with a lock? Generally would prefer to have some kind of API to interact with this. Also, should this be part of `WorkflowRegistry`?

saqadri: Better documentation on key/value would be great as well

src/mcp_agent/mcp/client_proxy.py
```diff
@@ -149,46 +149,109 @@ async def notify_via_proxy(
...
-            timeout_float = float(str(timeout_str).strip())
-        except Exception:
+            from temporalio import workflow, activity
+            in_temporal = workflow.in_workflow()
+            if in_temporal:
+                workflow_id = workflow.info().workflow_id
+            else:
+                in_temporal = activity.in_activity()
+                if in_temporal:
+                    workflow_id = activity.info().workflow_id
```
saqadri, line 171: This is unreachable code right?

src/mcp_agent/mcp/client_proxy.py
```diff
@@ -149,46 +149,109 @@ async def notify_via_proxy(
...
-    if timeout_str is None:
-        timeout_float = None  # no timeout by default; activity timeouts still apply
-    else:
+    if make_async_call:
+        # Make sure we're running in a Temporal workflow context
         try:
-            timeout_float = float(str(timeout_str).strip())
-        except Exception:
+            from temporalio import workflow, activity
+            in_temporal = workflow.in_workflow()
```
saqadri, line 165:
```suggestion
            in_workflow = workflow.in_workflow()
```

src/mcp_agent/mcp/sampling_handler.py
```diff
@@ -115,10 +185,14 @@ async def _human_approve_response(
 
         from mcp_agent.human_input.types import HumanInputRequest
 
+        response_summary = _format_sampling_response_for_human(result)
+
         req = HumanInputRequest(
             prompt=(
                 "LLM has generated a response. Respond 'approve' to send, "
                 "anything else to reject (your input will be recorded as reason)."
```
saqadri, line 193: Do we need this in the sampling handler btw? Or can it just send it without human intervention.

saqadri (commented):
Generally looks good, but concerned about the use of a global _workflow_states.
At minimum would be great to consolidate that into WorkflowRegistry, or have APIs to interact with it so callers aren't mutating the dict directly.

src/mcp_agent/executor/workflow.py
roman-van-der-krogt: True, but that is somehow out of scope here. Not sure by what logic exactly, but it errors out. (See also the import on line 456, where we import `workflow` yet again.)

src/mcp_agent/mcp/client_proxy.py
roman-van-der-krogt: No, it is not. When we are in a workflow and run `SystemActivities.relay_request` as an activity, when we hit this point, `workflow.in_workflow()` returns `False`, and `activity.in_activity()` returns `True`, and we can access the outer workflow id from `activity.info()`.

src/mcp_agent/mcp/sampling_handler.py
roman-van-der-krogt: According to the [spec](https://modelcontextprotocol.io/specification/2025-06-18/client/sampling#user-interaction-model):

```
Applications SHOULD:
- [...]
- Present generated responses for review before delivery
```

So, I think we should (in the lower case sense) keep this

src/mcp_agent/executor/temporal/session_proxy.py
```diff
@@ -117,20 +117,49 @@ async def request(
...
                 method,
                 params or {},
             )
+
+            if execution_info.get("error", "") != "":
+                return execution_info
+
+            signal_name = execution_info.get("signal_name", "")
+
+            if signal_name == "":
```
saqadri: nit: should we check for None as well?

src/mcp_agent/executor/temporal/session_proxy.py
```diff
@@ -248,17 +277,17 @@ async def list_roots(self) -> types.ListRootsResult:
...
-        *,
-        max_tokens: int,
-        system_prompt: str | None = None,
-        include_context: types.IncludeContext | None = None,
-        temperature: float | None = None,
-        stop_sequences: List[str] | None = None,
-        metadata: Dict[str, Any] | None = None,
-        model_preferences: types.ModelPreferences | None = None,
-        related_request_id: types.RequestId | None = None,
+            self,
```
saqadri: Just confirming this indentation is intentional

src/mcp_agent/workflows/factory.py
```diff
@@ -72,27 +72,24 @@ def create_llm(
...
-        provider=provider,
-        model=model,
-        request_params=request_params,
-        context=context,
-    )
 
-    return factory(agent=agent)
+@overload
+def create_llm(
+    agent: str,
```
saqadri: This should probably be agent_name

saqadri (approved):
Looks awesome! Just a few minor comments

src/mcp_agent/executor/temporal/session_proxy.py
roman-van-der-krogt: That's my formatter deciding it needs more indentiation :/

src/mcp_agent/workflows/factory.py
roman-van-der-krogt: Indeed. Good catch

PR #504, opened 2025-09-17
Make mcp-agent Context a derivative of FastMCP context

saqadri:
This makes it much easier to interface with the MCP server session everywhere that `Context` is available.

It's especially useful for @app.tool decorated functions:

@app.tool
def my_agent_fn(params: Any, app_ctx: Context) -> str:
    app_ctx.elicit(...)
    app_ctx.log(...)
    return "Mission accomplished"

### DIFF
--- a/src/mcp_agent/core/context.py
+++ b/src/mcp_agent/core/context.py
@@ -4,13 +4,14 @@ A central context object to store global state that is shared across the applica
 
 import asyncio
 import concurrent.futures
-from typing import Any, List, Optional, TYPE_CHECKING
+from typing import Any, List, Optional, TYPE_CHECKING, Literal
 import warnings
 
 from pydantic import BaseModel, ConfigDict
 
 from mcp import ServerSession
 from mcp.server.fastmcp import FastMCP
+from mcp.server.fastmcp import Context as MCPContext
 
 from opentelemetry import trace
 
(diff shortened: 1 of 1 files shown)


roman-van-der-krogt (approved):

src/mcp_agent/app.py
```diff
@@ -784,13 +845,25 @@ def decorator(fn: Callable[P, R]) -> Callable[P, R]:
                     "source_fn": fn,
                     "structured_output": structured_output,
                     "description": description or (fn.__doc__ or ""),
+                    "title": title,
+                    "annotations": annotations_obj,
+                    "icons": icons_list,
+                    "meta": meta_payload,
```
rholinshead, line 864: Should this be `_meta`?

saqadri: Hmmm I looked at the MCP SDK and it looks like the decorator has meta, not _meta. Where did you see _meta?

rholinshead: Ah, `_meta` is what's used by openai apps -- https://developers.openai.com/apps-sdk/build/mcp-server

Looking at the MCP docs it seems it may be abused for that reason -- https://github.com/modelcontextprotocol/modelcontextprotocol/blob/47339c03c143bb4ec01a26e721a1b8fe66634ebe/docs/specification/draft/basic/index.mdx#_meta -- since supposed to be reserved for MCP itself

saqadri: So `meta` on the `tool` decorator will be set as `_meta` on the actual tool registration in MCP. See https://github.com/modelcontextprotocol/python-sdk/pull/1463

However that change needs to land in pypi first so we can consume it

PR #505, opened 2025-09-18
Fix: executor signal timeout

rholinshead:
## Summary
Currently, `wait_for_signal` is not propagating the `timeout_seconds` to the underlying signal bus so waiting for signal will run indefinitely. To fix, just pass it through.

Also, fix timeout catch around temporal `workflow.wait_condition` since it raises `asyncio.Timeout` instead of the regular temporal workflow `Timeout` exception 

## Testing
```bash
make lint
make format
make tests
```

Updated the `basic_agent_server` example and tested `PauseResumeWorkflow` with a 10s timeout to ensure the run fails

## Summary by CodeRabbit

- New Features
  - Support for nested server execution during sampling and elicitation.
  - Optional timeouts when waiting for workflow signals.
  - Enhanced LLM integration and interactive elicitation via console.

- Improvements
  - More reliable pause/resume behavior with explicit timeout handling and clearer failure propagation.
  - Consistent timeout handling across executors and workflows.

- Tests
  - Added tests to validate signal timeout behavior and error surfacing.


saqadri (approved):
nice catch!

PR #507, opened 2025-09-18
add human_interaction handler using MCP elicitation

roman-van-der-krogt:
WIP, as it builds on the elicitation workflow via signal handler (PR #502).

- Added `human_input/elicitation_handler.py`, a human interaction handler that uses MCP elicitation to request user input. This builds on the existing elicitation handling.
- Added `examples/human_input/temporal` that shows how to use it in an MCP app, including a README with more details.
- Renamed the existing `human_input/handler.py` to `human_input/console_handler.py` for clarity


src/mcp_agent/executor/temporal/session_proxy.py
```diff
@@ -189,20 +118,35 @@ async def request(
...
 
-            # Wait for the _elicitation_response signal to be triggered
-            await _twf.wait_condition(
-                lambda: has_signal_response_sync(exec_id)
+            if execution_info.get("error", "") != "":
+                return execution_info
+
+            signal_name = execution_info.get("signal_name", "")
+
+            if signal_name == "":
```
saqadri: Same question as in other PR if we need to check for None

PR #511, opened 2025-09-18
feat: bundle comments and file ignore

alienzach:
# Add support for ignore files in deploy bundling

## TL;DR

Added support for gitignore-style pattern matching during deployment to exclude files from being bundled.

## What changed?

- Added two new CLI options to the `deploy` command:
    - `--ignore`: Uses `.mcpacignore` in the current directory
    - `--ignore-file <path>`: Specifies a custom ignore file path
- Created a new `bundle_utils.py` module with utilities for:
    - Parsing gitignore-style patterns into a PathSpec matcher
    - Filtering files during the bundling process
- Enhanced the bundling process to respect ignore patterns while preserving default excludes
- Updated documentation with examples and usage information

## How to use?

```
# Use default .mcpacignore in current directory
mcp-agent deploy my-agent --ignore

# Use a custom ignore file
mcp-agent deploy my-agent --ignore-file .deployignore

# Both options can be combined (--ignore-file takes precedence)
mcp-agent deploy my-agent --ignore --ignore-file .customignore
```

Patterns follow gitignore syntax and are evaluated relative to the project directory.

## Why make this change?

This change gives users more control over which files are included in their deployment bundles. It allows:

- Excluding large files or directories that aren't needed for deployment
- Preventing sensitive files from being bundled
- Following standard gitignore patterns that developers are already familiar with
- Customizing ignore rules per project or deployment

## Summary by CodeRabbit

- New Features
  - Deploy supports gitignore-style exclusions via an ignore file; preserves default excludes when none provided.
  - Added --ignore-file <path> option with auto-detect of .mcpacignore; explicit --ignore-file takes precedence.
  - Deployment now logs bundled project files (excluding internal helpers) before packaging.
- Documentation
  - CLI reference and cloud overview updated with ignore-file behavior, precedence, and examples.
- Tests
  - Expanded tests for ignore-file detection, precedence, pathspec-driven ignores, and secret handling.
- Chores
  - Added optional dependency pathspec>=0.12.1 for ignore pattern support.

### DIFF
--- a/docs/cli-reference.mdx
+++ b/docs/cli-reference.mdx
@@ -272,8 +272,7 @@ mcp-agent deploy my-agent --ignore-file .deployignore
 
 Ignore file usage
 
-- If `.mcpacignore` exists in the working (or `--config-dir`) directory, it is applied automatically.
-- `--ignore-file <path>` overrides the default detection when you need a different ignore file.
+- Detection order: 1) `--ignore-file <path>` (highest precedence), 2) `.mcpacignore` in `--config-dir`, 3) `.mcpacignore` in the working directory (CWD).
 - Patterns follow gitignore syntax and are evaluated relative to `--config-dir`.
 
 ### mcp-agent cloud configure
--- a/src/mcp_agent/cli/cloud/commands/deploy/bundle_utils.py
+++ b/src/mcp_agent/cli/cloud/commands/deploy/bundle_utils.py
@@ -62,17 +62,19 @@ def should_ignore_by_gitignore(
         full_path = current_path / name
         try:
             rel_path = full_path.relative_to(project_dir)
-            # Normalize to POSIX separators so patterns work cross-platform (Windows too)
-            rel_path_str = rel_path.as_posix()
-
-            # Match files exactly; for directories also try with a trailing slash
-            # to respect patterns like `build/`.
-            if spec.match_file(rel_path_str):
-                ignored.add(name)
-            elif full_path.is_dir() and spec.match_file(rel_path_str + "/"):
-                ignored.add(name)
         except ValueError:
             # If `full_path` is not under `project_dir`, ignore matching is skipped.
             continue
+
+        # Normalize to POSIX separators so patterns work cross-platform (Windows too)
+        rel_path_str = rel_path.as_posix()
+
+        # Match files exactly; for directories also try with a trailing slash
+        # to respect patterns like `build/`.
+        if spec.match_file(rel_path_str):
+            ignored.add(name)
+        elif full_path.is_dir() and spec.match_file(rel_path_str + "/"):
+            ignored.add(name)
+
 
     return ignored
--- a/src/mcp_agent/cli/cloud/commands/deploy/main.py
+++ b/src/mcp_agent/cli/cloud/commands/deploy/main.py
@@ -119,7 +119,10 @@ def deploy_config(
     ignore_file: Optional[Path] = typer.Option(
         None,
         "--ignore-file",
-        help="Path to ignore file (gitignore syntax). Defaults to '.mcpacignore' if present.",
+        help=(
+            "Path to ignore file (gitignore syntax). Precedence: 1) --ignore-file <path>, "
+            "2) .mcpacignore in --config-dir, 3) .mcpacignore in working directory."
+        ),
         exists=False,
         readable=True,
         dir_okay=False,
--- a/docs/cloud/overview.mdx
+++ b/docs/cloud/overview.mdx
@@ -24,7 +24,7 @@ description: "[In beta] Deploy and manage AI agents as MCP servers"
 
 <Note>When packaging your mcp-agent app for the cloud, our CLI will be searching for `main.py`. The entire directory will be deployed, so you can reference other files and upload other assets.</Note>
 
-<Note>You can optionally exclude files from the bundle using an ignore file (gitignore syntax). If `.mcpacignore` exists in your project directory it is applied automatically, or pass `--ignore-file <path>` to point to a different file.</Note>
+<Note>You can optionally exclude files from the bundle using an ignore file (gitignore syntax). Precedence: 1) `--ignore-file <path>` (explicit override), 2) `.mcpacignore` in `--config-dir`, 3) `.mcpacignore` in the working directory (CWD).</Note>
 
 2. Make sure you have either a `pyproject.toml` or a `requirements.txt` file in your app directory.
 


alienzach:
* **#511** <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/511?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/511?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/mcp_agent/cli/cloud/commands/deploy/bundle_utils.py
```diff
@@ -0,0 +1,328 @@
...
+        # Dump it back without comments
+        cleaned = yaml.dump(data, default_flow_style=False, sort_keys=False, allow_unicode=True)
+
+        return cleaned
+    except yaml.YAMLError:
+        # If parsing fails, return original content
+        return yaml_content
+
+
+def clean_python_comments(python_content: str) -> str:
```
rholinshead: Is there any value to removing comments from their code? It seems to just add unnecessary complexity / potential for errors

src/mcp_agent/cli/cloud/commands/deploy/bundle_utils.py
```diff
@@ -0,0 +1,328 @@
...
+                    f.write(cleaned_content)
+                cleaned_count += 1
+
+        except Exception:
+            continue
+
+    return cleaned_count
+
+
+def clean_python_files_in_directory(directory: Path) -> int:
```
rholinshead: I think we should not do this due to the added complexity/potential for errors

src/mcp_agent/cli/cloud/commands/deploy/bundle_utils.py
```diff
@@ -0,0 +1,328 @@
...
+            elif full_path.is_dir() and spec.match_file(rel_path_str + '/'):
+                ignored.add(name)
+        except ValueError:
+            # Path is not relative to project_dir, don't ignore
+            continue
+
+    return ignored
+
+
+def clean_yaml_comments(yaml_content: str) -> str:
```
rholinshead: There is some risk to this in the sense that users deploying their app project may expect the files to be unchanged. Consider the following scenario which would be broken by this:
1. I have an app with some yaml files for different settings / configurations and have comments in the file to explain some properties
2. In my app I ask an LLM to use some subset of yaml configuration files for doing some action; I expect the comments to exist and give the necessary context
3. The LLM reads the file but no comments exist with context, resulting in poor performance vs running it locally

src/mcp_agent/cli/cloud/commands/deploy/wrangler_wrapper.py
```diff
@@ -156,22 +162,43 @@ def wrangler_deploy(app_id: str, api_key: str, project_dir: Path) -> None:
...
+            # Keep existing hardcoded exclusions (highest priority)
             for name in names:
                 if (name.startswith(".") and name not in {".env"}) or name in {
                     "logs",
                     "__pycache__",
                     "node_modules",
                     "venv",
-                    MCP_SECRETS_FILENAME,
+                    MCP_SECRETS_FILENAME,  # Continue excluding mcp_agent.secrets.yaml
+                    f"{MCP_SECRETS_FILENAME}.example",  # Exclude mcp_agent.secrets.yaml.example
```
rholinshead: I don't think there's any need to exclude this example file. It may even be useful to keep so that it's clearer what api keys are expected if we ever need to analyze the bundle for that

src/mcp_agent/cli/cloud/commands/deploy/wrangler_wrapper.py
```diff
@@ -156,22 +162,43 @@ def wrangler_deploy(app_id: str, api_key: str, project_dir: Path) -> None:
...
+            gitignore_ignored = should_ignore_by_gitignore(path_str, names, project_dir, gitignore_spec)
+            ignored.update(gitignore_ignored)
+
             return ignored
 
         shutil.copytree(project_dir, temp_project_dir, ignore=ignore_patterns)
 
+        # Clean comments from YAML and Python files to prevent secret leakage
+        yaml_cleaned = clean_yaml_files_in_directory(temp_project_dir)
+        python_cleaned = clean_python_files_in_directory(temp_project_dir)
```
rholinshead: I think there's probably a very low probability of there being any secrets in the comments of python files. And probably likewise for yaml files?

alienzach (commented):
Tested locally with examples and working as expected.

docs/cli-reference.mdx
```diff
@@ -251,6 +251,8 @@ mcp-agent deploy <APP_NAME> [OPTIONS]
 | `--dry-run` | Validate without deploying | `false` |
 | `--api-url` | API base URL | `https://deployments.mcp-agent.com` |
 | `--api-key` | API key | From env or config |
+| `--ignore` | Use `.mcpacignore` in current directory to ignore files | `false` |
```
rholinshead: Fwiw I think it would be better to do this by default if there's a `.mcpacignore` in the dir and then we only need a single `--ignore` command pointing to a custom file

src/mcp_agent/cli/cloud/commands/deploy/main.py
```diff
@@ -104,6 +104,21 @@ def deploy_config(
...
+    ignore_file: Optional[Path] = typer.Option(
+        None,
+        "--ignore-file",
+        help="Specify custom ignore file (gitignore syntax)",
+        exists=False,
+        readable=True,
+        dir_okay=False,
+        file_okay=True,
+        resolve_path=True,
+    ),
```
rholinshead: See comment above; it seems weird to have two ignore-related commands. If there's a `.mcpacignore` file and no ignore param used it seems weird to not use that file (.gitignore is automatic)

src/mcp_agent/cli/cloud/commands/deploy/wrangler_wrapper.py
```diff
@@ -156,18 +162,41 @@ def wrangler_deploy(app_id: str, api_key: str, project_dir: Path) -> None:
...
+            print_info("No ignore file provided; applying default excludes only")
+
+        # Copy the entire project to temp directory, excluding unwanted directories and the live secrets file
+        def ignore_patterns(path_str, names):
             ignored = set()
+
+            # Keep existing hardcoded exclusions (highest priority)
             for name in names:
                 if (name.startswith(".") and name not in {".env"}) or name in {
                     "logs",
```
rholinshead, line 196: I guess we _are_ already excluding logs dir? One thing I remember from the sync was using the logs dir parsed from the config, if that can be done instead (maybe a follow up if it's too much to include here)

src/mcp_agent/cli/cloud/commands/deploy/wrangler_wrapper.py
```diff
@@ -201,6 +230,34 @@ def ignore_patterns(_path, names):
...
+        added_deployment_files = []
+        for root, _dirs, files in os.walk(temp_project_dir):
+            for filename in files:
+                rel = Path(root).relative_to(temp_project_dir) / filename
+                # Map transient .mcpac.py back to the original filename for logging
+                if filename.endswith(".mcpac.py"):
+                    orig_rel = str(rel)[: -len(".mcpac.py")]
+                    bundled_original_files.append(orig_rel)
+                elif filename in {"wrangler.toml", "mcp_deploy_breadcrumb.py"}:
+                    added_deployment_files.append(str(rel))
```
rholinshead: I don't think the end user should need to know/care about these (implementation details)

tests/cli/commands/test_deploy_command.py
```diff
@@ -74,6 +74,9 @@ def test_deploy_command_help(runner):
     assert "--api-url" in clean_text
     assert "--api-key" in clean_text
     assert "--non-interactive" in clean_text
+    # New ignore options
```
rholinshead:
```suggestion
```

rholinshead (commented):
Looks good in general but a couple things:
- from UX perspective, seems weird to have two different `ignore` options; I would recommend just the one for overriding ignore file path and otherwise use `.mcpacignore` if it exists in the project dir
- I don't think we want to tell the user about any 'internal' files included in the deployment since they are implementation details and will just be confusing

Also a comment on the logs dir handling, which may make more sense in a followup

src/mcp_agent/cli/cloud/commands/deploy/wrangler_wrapper.py
```diff
@@ -201,6 +230,26 @@ def ignore_patterns(_path, names):
...
+                    orig_rel = str(rel)[: -len(".mcpac.py")]
+                    bundled_original_files.append(orig_rel)
+                else:
+                    bundled_original_files.append(str(rel))
+
+        bundled_original_files.sort()
+        if bundled_original_files:
+            print_info(f"Bundling {len(bundled_original_files)} project file(s):")
+            for p in bundled_original_files:
+                print_info(f" - {p}")
```
rholinshead: Will this print out with "INFO: " prefix? If so, probably better to just use `console.print`

rholinshead (approved):
Just one final comment on the printing of the bundled files, to avoid a string of "Info: " -prefixed logs

PR #512, opened 2025-09-19
fix: v26 tests

alienzach:


.github/workflows/checks.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         with:
           python-version-file: ".python-version"
 
-      - name: Install dependencies
-        run: make sync
+      - name: Install the project (frozen)
+        run: uv sync --frozen --all-extras --dev
```
saqadri: Why this change?

src/mcp_agent/config.py
```diff
@@ -783,10 +783,10 @@ def deep_merge(base: dict, update: dict) -> dict:
             yaml_settings = yaml.safe_load(buf) or {}
 
             # Preload is authoritative: construct from YAML directly (no env overlay)
-            settings = Settings(**yaml_settings)
-            if set_global:
-                _set_and_warn_global_settings(settings)
-            return settings
+            # settings = Settings(**yaml_settings)
```
saqadri: I don't understand why this is being commented out

saqadri (changes requested):
This change doesn't seem quite right to me. Can you please elaborate what the intended behavior is?

I also don't like commented out code without a clear documented reason for it being commented out

alienzach:
@saqadri this PR is in draft. it's a WIP to debug what's going on. it is used to rretrigger tests in CI. please do not merge.

saqadri:
> @saqadri this PR is in draft. it's a WIP to debug what's going on. it is used to rretrigger tests in CI. please do not merge. 

Ah sorry about that, I didn't see that @alienzach!

alienzach:
This PR is a second attempt for #501

src/mcp_agent/config.py
```diff
@@ -767,7 +808,10 @@ def deep_merge(base: dict, update: dict) -> dict:
             yaml_settings = yaml.safe_load(buf) or {}
 
             # Preload is authoritative: construct from YAML directly (no env overlay)
-            return Settings(**yaml_settings)
+            settings = Settings(**yaml_settings)
+            if set_global:
+                _set_and_warn_global_settings(settings)
```
saqadri: I don't know if we should be setting global in this case

src/mcp_agent/config.py
```diff
@@ -6,14 +6,14 @@
 import sys
 from io import StringIO
 from pathlib import Path
-from typing import Dict, List, Literal, Optional, Set
+from typing import Dict, List, Literal, Optional, Set, Union
```
saqadri: I prefer using the `T | R` notation instead of `Union`

src/mcp_agent/config.py
```diff
@@ -782,52 +826,44 @@ def deep_merge(base: dict, update: dict) -> dict:
...
-        if not config_file.exists():
+        if not _check_file_exists(config_file):
             raise FileNotFoundError(f"Config file not found: {config_path}")
     else:
         config_file = Settings.find_config()
 
     # If we found a config file, load it
-    if config_file and config_file.exists():
-        with open(config_file, "r", encoding="utf-8") as f:
-            yaml_settings = yaml.safe_load(f) or {}
```
saqadri, line 793: Why this change?

alienzach: This is to isolate file operations out from the core logic, so that they can be mocked in test without mocking system like file ops. See: https://github.com/lastmile-ai/mcp-agent/pull/512/commits/ab443cd5f165ed3d38e15615314a9f62d8e815a2

src/mcp_agent/config.py
alienzach: updated.

src/mcp_agent/config.py
alienzach: we can remove the global state set here. however i might be confusing myself by reading your previous comment: https://github.com/lastmile-ai/mcp-agent/pull/512#discussion_r2361732172

PR #515, opened 2025-09-19
Fix/revert to working main

rholinshead:
## Description

git revert 7d0f9db6e5b12ef11100e5d632551d12460f0cc5..HEAD

to get main back into working order while we figure out the test issues


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "mcp-agent"
-version = "0.1.26"
+version = "0.1.25"
```
saqadri, line 3: Has .26 been published?

saqadri (approved):

pyproject.toml
rholinshead: No, it failed publishing due to the test failure (despite the version bump PR passing)

saqadri: Sounds great, thank you!

PR #518, opened 2025-09-21
Temporal plugin implementation

StreetLamb:


src/mcp_agent/executor/temporal/plugin.py
```diff
@@ -0,0 +1,270 @@
...
+        return self.next_worker_plugin.configure_worker(config)
+
+    async def run_worker(self, worker: Worker) -> None:
+        """Run the worker with MCP Agent context."""
+        # Set up any necessary context before running the worker
+        await self.next_worker_plugin.run_worker(worker)
+
+    def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:
+        """Configure the replayer with MCP Agent settings."""
+        config["data_converter"] = self._get_new_data_converter(
```
drewhoskins-temporal: I believe you need to add `ContextPropagationInterceptor`  here as well, and potentially `workflow_runner`.  
It might be worth setting up a test that runs `Replayer` so that users who are doing replay tests can do them.

StreetLamb: Thank you for the feedback! Updated `configure_replayer` to configure workflow_runner, workflows and interceptors as well.

src/mcp_agent/executor/temporal/plugin.py
```diff
@@ -0,0 +1,342 @@
...
+from mcp_agent.executor.temporal.session_proxy import SessionProxy
+from mcp_agent.logging.logger import get_logger
+
+if TYPE_CHECKING:
+    from mcp_agent.app import MCPApp
+
+logger = get_logger(__name__)
+
+
+class MCPAgentPlugin(ClientPlugin, WorkerPlugin):
```
drewhoskins-temporal: FYI we are working on a simpler plugin interface here: https://github.com/temporalio/sdk-python/pull/1139/files?diff=split&w=1

pyproject.toml
```diff
@@ -36,7 +36,7 @@ dependencies = [
 
 [project.optional-dependencies]
 temporal = [
-    "temporalio[opentelemetry]>=1.10.0",
+    "temporalio[opentelemetry]",
```
StreetLamb, line 39: Update this once the Temporal's SimplePlugin feature is released.

PR #527, opened 2025-09-29
Allow specifying app name optionally in deploy command

saqadri:
Makes the app name optional in the deploy command. In case it's not specified, we now look in the config to see if an app name is defined there, and if not, default to `"default"` as the name.

Also improved the wording for the secrets prompt. 

## Summary by CodeRabbit

* New Features
  * Configure app name and description in settings; app uses these values when available.
  * Add a "Default Model" option for Google settings.
  * Deploy command supports working_dir and optional config_dir, and auto-defaults app name/description from config.

* Refactor
  * Simplified OpenTelemetry exporter config to a string list (console, file, otlp); clarified OTLP settings requirement.

* Tests
  * Added deploy tests for config-driven defaults and updated test formatting/fixtures.

### DIFF
--- a/tests/cli/commands/test_deploy_command.py
+++ b/tests/cli/commands/test_deploy_command.py
@@ -327,6 +327,67 @@ def test_deploy_uses_config_description_when_not_provided(runner, temp_config_di
     assert create_call.kwargs["description"] == "Configured app description"
 
 
+def test_deploy_uses_defaults_when_config_cannot_be_loaded(runner, temp_config_dir):
+    """If config parsing fails, fall back to default name and unset description."""
+
+    config_path = temp_config_dir / MCP_CONFIG_FILENAME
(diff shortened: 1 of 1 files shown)


schema/mcp-agent.config.schema.json
```diff
@@ -774,95 +786,6 @@
       "title": "MCPSettings",
       "type": "object"
     },
-    "OTelConsoleExporterSettings": {
```
rholinshead, line 777: Unrelated, but I suppose we can also clean up the fallback handling for the deprecated settings now?

rholinshead:
Thanks!

schema/mcp-agent.config.schema.json
saqadri: Yes will do

PR #528, opened 2025-09-29
Fix bundle path for .mcpac.py files on Windows

rholinshead:
## Description
The following error is seen when trying to deploy on Windows:

Screenshot 2025-09-29 at 1 54 20 PM

I believe the issues is due to our f-string use in constructing the path, so fix to use Path methods only

## Testing
```bash
make lint
make format
make tests
```

```bash
uv run mcp-agent deploy MCPAgentServerTemporal -c examples/mcp_agent_server/temporal 
```
Deployed successfully

### DIFF
--- a/src/mcp_agent/cli/cloud/commands/deploy/wrangler_wrapper.py
+++ b/src/mcp_agent/cli/cloud/commands/deploy/wrangler_wrapper.py
@@ -195,8 +195,7 @@ def wrangler_deploy(app_id: str, api_key: str, project_dir: Path) -> None:
                     continue
 
                 # For non-Python files, rename with .mcpac.py extension to be included as py files
-                relative_path = file_path.relative_to(temp_project_dir)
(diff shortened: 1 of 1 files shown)


src/mcp_agent/cli/cloud/commands/deploy/wrangler_wrapper.py
```diff
@@ -195,8 +195,7 @@ def ignore_patterns(_path, names):
                     continue
 
                 # For non-Python files, rename with .mcpac.py extension to be included as py files
-                relative_path = file_path.relative_to(temp_project_dir)
-                py_path = temp_project_dir / f"{relative_path}.mcpac.py"
+                py_path = file_path.with_suffix(file_path.suffix + ".mcpac.py")
```
rholinshead, line 198: File is already in `temp_project_dir` so no need to do any relative path logic

saqadri (approved):

PR #547, opened 2025-10-08
OAuth support for mcp-agent

saqadri:
Various fixes for full OAuth 2.1 support to mcp-agent, covering both inbound server authorization (resource metadata, token verification) and outbound delegated authorization to downstream MCP servers. 

## Overall:

- Introduced a new mcp_agent.oauth subsystem, including token records, pluggable stores (in-memory + optional Redis), the TokenManager, metadata discovery helpers, and an AuthorizationFlowCoordinator that drives interactive OAuth flows via auth/request.

- Extended MCPApp/context initialization to support optional session ID overrides and to bootstrap OAuth token management based on configuration.
Preconfigured tokens are captured once and stored safely; session-aware fallbacks are kept intact.

- Hardened the server side (app_server) to surface `/internal/oauth/callback`, handle workflow pre-auth correctly, and attach authenticated identities without leaking synthetic ones.

- Added tests for URL normalization, metadata fallbacks, token- management behaviour, and the new session-ID override.

## Issues fixed:

- OAuth metadata discovery building the wrong URLs. For something like https://api.githubcopilot.com/ mcp, this produces `https://api.githubcopilot.com/.well-known/oauth-protected-resource/mcp`, which is incorrect and will 404 against real providers. The spec expects `scheme://host/.well-known/...` regardless of path.

- Preconfigured-token user identity broke invalidation
  - _set_upstream... in src/mcp_agent/server/app_server.py:178 sets context.current_user to a synthetic identity derived from create_default_user_for_preconfigured_tokens(session_id). When OAuthHttpxAuth later tries to invalidate() after a 401, it uses that session-specific identity. But the stored token uses the session-less synthetic identity, so the cache entry is never deleted. Resulting in keep retrying a dead token forever.

- workflow_pre_auth assumes an authenticated user


roman-van-der-krogt:
Haven't looked at the code yet, so maybe this comment is void, but the issue "OAuth metadata discovery building the wrong URLs" is incorrect. This is working as per the RFC (https://datatracker.ietf.org/doc/html/rfc9728#section-3.1):

> The consumer of the metadata would make the following request when the resource identifier is https://resource.example.com/resource1 and the well-known URI path suffix is oauth-protected-resource to obtain the metadata, since the resource identifier contains a path component:
> 
>   GET /.well-known/oauth-protected-resource/resource1 HTTP/1.1
>   Host: resource.example.com
> 

(This can be checked by noticing that [https://api.githubcopilot.com/.well-known/oauth-protected-resource/mcp](https://api.githubcopilot.com/.well-known/oauth-protected-resource/mcp) is where the metadata is hosted and does not result in a 404, whereas [https://api.githubcopilot.com/.well-known/oauth-protected-resource](https://api.githubcopilot.com/.well-known/oauth-protected-resource) does result in a page not found error. Directly contradicting the "fix".)

docs/advanced/monitoring.mdx
```diff
@@ -1460,7 +1460,7 @@ receivers:
 - name: 'critical-alerts'
   slack_configs:
   - channel: '#critical-alerts'
-    title: '=� CRITICAL: MCP Agent Alert'
+    title: '=� CRITICAL: MCP Agent Alert'
```
roman-van-der-krogt, line 1463: Are these characters expected? `ï¿½`

src/mcp_agent/oauth/manager.py
```diff
@@ -35,18 +40,94 @@
...
+
+def _candidate_authorization_metadata_urls(
+    parsed_authorization_server: URL,
+) -> list[str]:
+    base = parsed_authorization_server.copy_with(path="", query=None, fragment=None)
+    path = parsed_authorization_server.path.lstrip("/")
+    candidates = []
+    if path:
+        candidates.append(
+            str(base.copy_with(path=f"/.well-known/oauth-authorization-server/{path}"))
```
roman-van-der-krogt, line 119: First candidate is per the RFC 👍

src/mcp_agent/server/app_server.py
```diff
@@ -1574,8 +1557,20 @@ async def workflow_pre_auth(
...
-            raise ToolError("Token storage not available.")
+        if not app_context.token_manager:
+            raise ToolError("OAuth token manager not available.")
+
+        current_user = app_context.current_user
+        if current_user is None:
+            session_id = getattr(app_context, "session_id", None)
+            if session_id:
+                current_user = OAuthUserIdentity(
+                    provider="mcp-session", subject=str(session_id)
```
roman-van-der-krogt: The `session_id` in the context is initialised when the server starts and then remains the same. So two users will definitely interfere with each other here. This is why I tried reading the session id from the request parameters. These would be unique

> ```
> session_id = ctx.request_context.request.query_params.get("session_id")
> identity = create_default_user_for_preconfigured_tokens(session_id)
> ```

src/mcp_agent/oauth/manager.py
```diff
@@ -237,14 +437,40 @@ async def invalidate(
...
+            default_key = self._build_store_key(
+                self._default_identity,
+                canonical_resource,
+                canonical_auth_server,
+                tuple(scopes),
+            )
+            await self._token_store.delete(default_key)
+        if session_id:
+            session_identity = OAuthUserIdentity(
+                provider="mcp-session", subject=str(session_id)
```
roman-van-der-krogt: See also my comment below on `app_server`, this depends on the `session_id` being different from one user to another

src/mcp_agent/app.py
```diff
@@ -333,6 +387,52 @@ async def initialize(self):
             },
         )
 
+    async def _initialize_preconfigured_tokens(self, token_manager):
```
roman-van-der-krogt: Do we want to keep supporting this, seeing as it's undocumented?

PR #548, opened 2025-10-08
Add better identity management in app server

saqadri:
## Problems

- Identity Leakage: The server used to stash the authenticated user in app.context.current_user, a shared, long-lived object. Concurrent requests (or workflows coming back later) would overwrite that field, so TokenManager might refresh or invalidate tokens for the wrong human.

- Temporal Callbacks Missing Identity: Activities and workflows emit `internal/session/...` calls from separate processes; those relays ran without any authenticated context, causing delegated OAuth flows (auth/request) to fail because the TokenManager saw no user identity.

- TokenManager Context Dependence: Downstream OAuth clients relied exclusively on context.current_user and context.session_id. If either was missing (e.g., because we intentionally avoided touching shared context), they couldn’t fetch or refresh tokens.

## Changes

We introduce per-request/per-execution identity scoping (via contextvars) and memo propagation so each outbound OAuth interaction — and every Temporal callback—runs under the correct user without mutating global state.

### DIFF
--- a/src/mcp_agent/executor/workflow.py
+++ b/src/mcp_agent/executor/workflow.py
@@ -14,6 +14,8 @@ from typing import (
 )
 
 
+from contextvars import Token
+
 from pydantic import BaseModel, ConfigDict, Field
 from mcp_agent.core.context_dependent import ContextDependent
 from mcp_agent.executor.workflow_signal import (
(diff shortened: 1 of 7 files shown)


saqadri:
We're going in a different direction. All updates in https://github.com/lastmile-ai/mcp-agent/tree/feat/mcp_agent_oauth2

PR #550, opened 2025-10-08
[rfc] fuzzy matching to exporter schema

saqadri:
Ensure OpenTelemetrySettings.exporters tolerate list items that drop the type, inferring it when possible and emitting a clear error otherwise.

Example:
base config (mcp_agent.config.yaml):
```
        otel:
          exporters:
            - type: otlp
              endpoint: "https://collector.default/v1/traces"
```

overlay (mcp_agent.secrets.yaml):
```
        otel:
          exporters:
            - headers:
                Authorization: !developer_secret OTEL_AUTH
```

merged result (what we load):
````
        otel:
          exporters:
            - type: otlp
              endpoint: "https://collector.default/v1/traces"
              headers:
                Authorization: !developer_secret OTEL_AUTH
```

also mixed entries like `["console", {"type": "file"}]` are handled fine:

During the before validator we normalize the list. Strings stay as strings; dict entries keep or infer their type.
In the `after` validator we walk the list: literal strings become ConsoleExporterSettings, and dict entries model-validate into their typed counterparts.

So a config like:
```
  otel:
    exporters:
      - console
      - { type: file, path: "traces.jsonl" }
```
will end up as the expected mix of typed exporters at runtime.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* New Features
  * Telemetry exporter configuration is now more flexible: the exporter type can be inferred automatically when omitted.
  * Improved backward compatibility: legacy exporter fields are accepted and converted into the new typed format.
  * Clearer validation: helpful errors are shown when the exporter type cannot be determined.

* Tests
  * Added comprehensive tests covering type inference for OTLP and File exporters, preservation of provided settings, and error handling for unrecognized configurations.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/src/mcp_agent/config.py
+++ b/src/mcp_agent/config.py
@@ -605,6 +605,24 @@ class OpenTelemetrySettings(BaseModel):
 
     model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
 
+    @staticmethod
+    def _guess_exporter_type(entry: Dict[str, Any]) -> str | None:
+        """
+        Infer an exporter type when an entry is missing the explicit `type`.
+
+        This helps when secrets overlays include only secret-bearing fields and
+        rely on the base config to supply the exporter type.
+        """
+
+        otlp_keys = {"endpoint", "headers", "compression", "certificate_file"}
+        file_keys = {"path", "path_settings"}
+
+        if any(key in entry for key in otlp_keys):
+            return "otlp"
+        if any(key in entry for key in file_keys):
+            return "file"
+        return None
+
     @model_validator(mode="before")
     @classmethod
     def _coerce_exporters_schema(cls, data: Dict) -> Dict:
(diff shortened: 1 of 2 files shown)


src/mcp_agent/config.py
alienzach, line 563: i think the root cause of issue is here. when we have "type: console", for example:
1. because "console" is the default value of type, when we deserialize settings object from base setting yaml with "ignore defaults", this ConsoleExporterSettings will be deserialzied to empty object {}.
2. then, we will use the yaml merge to merge the secrets volume into this base settings object, Pydantic validator will get an {} empty and Pydantic will fail at validation because without "type", it cannot decide which type that is.

saqadri:
Closing this PR in favor of #551

PR #551, opened 2025-10-09
refactor: otel settings have stable (de)serialization

alienzach:
# Refactored OpenTelemetry Exporter Configuration

### TL;DR

Refactored OpenTelemetry exporter configuration to support three schema versions with improved backward compatibility.

### What changed?

- Removed the discriminated union pattern with `type` field in favor of a more flexible approach
- Implemented three schema versions for exporter configuration:
    - V1: String exporters with legacy fields (`exporters: ["console", "file"]`)
    - V2: Discriminated union with type field (`exporters: [{type: "file", path: "/tmp/out"}]`)
    - V3: Dict key as discriminator (`exporters: [console: {}, file: {path: "trace.jsonl"}]`)
- Enhanced validation to properly handle all three formats with appropriate error messages
- Improved handling of legacy fields (`path`, `path_settings`, `otlp_settings`) for backward compatibility
- Added comprehensive tests for all schema versions and mixed configurations

### How to test?

1. Test with V1 string exporters:

```
otel:
  enabled: true
  exporters: ["console", "file", "otlp"]
  path: "/tmp/trace.jsonl"
  otlp_settings:
    endpoint: "http://collector:4318/v1/traces"
```

2. Test with V2 type-discriminated exporters:

```
otel:
  enabled: true
  exporters:
    - type: "console"
    - type: "file"
      path: "/tmp/trace.jsonl"
```

3. Test with V3 key-discriminated exporters:

```
otel:
  enabled: true
  exporters:
    - console: {}
    - file:
        path: "/tmp/trace.jsonl"
    - otlp:
        endpoint: "http://collector:4318/v1/traces"
```

4. Run the test suite: `pytest tests/test_config_exporters.py`

### Why make this change?

The previous configuration approach using discriminated unions with a `type` field was overly restrictive and not as user-friendly. This refactoring:

1. Makes the configuration more intuitive and flexible for users
2. Maintains backward compatibility with existing configurations
3. Allows for multiple exporters of the same type with different configurations
4. Simplifies the internal model by removing unnecessary inheritance
5. Provides clearer documentation and error messages for configuration issues

## Summary by CodeRabbit

- **New Features**
  - More flexible OpenTelemetry exporters: exporters can be simple names or named exporter blocks (e.g., otlp with endpoint/headers); supports multiple config sources.

- **Documentation**
  - Updated tracing READMEs and example YAMLs to reflect the new exporters structure and OTLP guidance (endpoints/authorization).

- **Refactor**
  - Normalized exporter schema/merging and removed legacy otlp_settings from defaults.

- **Tests**
  - Expanded validation and integration tests for mixed-format exporter configs and error cases.

### DIFF
--- a/src/mcp_agent/config.py
+++ b/src/mcp_agent/config.py
@@ -534,6 +534,7 @@ class TraceOTLPSettings(BaseModel):
 
     model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
 
+
 class ConsoleExporterSettings(BaseModel):
     """Console exporter uses stdout; no extra settings required."""
 
@@ -542,6 +543,7 @@ class ConsoleExporterSettings(BaseModel):
 
 class FileExporterSettings(BaseModel):
     """File exporter settings for writing traces to a file."""
+
     path: str | None = None
     path_settings: TracePathSettings | None = None
 
@@ -638,7 +640,10 @@ class OpenTelemetrySettings(BaseModel):
 
             # Handle BaseModel instances passed directly (e.g., from tests or re-validation)
             # If already a typed exporter settings instance, keep as-is (already finalized)
-            if isinstance(entry, (ConsoleExporterSettings, FileExporterSettings, OTLPExporterSettings)):
+            if isinstance(
+                entry,
+                (ConsoleExporterSettings, FileExporterSettings, OTLPExporterSettings),
+            ):
                 normalized.append(entry)
                 continue
 
@@ -663,8 +668,8 @@ class OpenTelemetrySettings(BaseModel):
 
             raise ValueError(
                 "OpenTelemetry exporters must be strings, type-tagged dicts, or "
-                "keyed mappings (e.g. `- console`, `- {type: \"file\"}`, "
-                "`- {file: {path: \"trace.jsonl\"}}`)."
+                'keyed mappings (e.g. `- console`, `- {type: "file"}`, '
+                '`- {file: {path: "trace.jsonl"}}`).'
             )
 
         data["exporters"] = normalized
@@ -707,7 +712,9 @@ class OpenTelemetrySettings(BaseModel):
             legacy_path_settings, TracePathSettings
         ):
             legacy_path_settings = TracePathSettings.model_validate(
-                getattr(legacy_path_settings, "model_dump", lambda **_: legacy_path_settings)()
+                getattr(
+                    legacy_path_settings, "model_dump", lambda **_: legacy_path_settings
+                )()
             )
 
         # Extract legacy otlp_settings and normalize to dict
@@ -749,27 +756,36 @@ class OpenTelemetrySettings(BaseModel):
                     payload = payload.model_dump(exclude_none=True)
                 elif not isinstance(payload, dict):
                     raise ValueError(
-                        "Exporter configuration must be a dict. Example: `- file: {path: \"trace.jsonl\"}`"
+                        'Exporter configuration must be a dict. Example: `- file: {path: "trace.jsonl"}`'
                     )
             else:
                 raise TypeError(f"Unexpected exporter entry: {exporter!r}")
 
             if exporter_name == "console":
                 console_settings = ConsoleExporterSettings.model_validate(payload or {})
-                finalized_exporters.append({"console": console_settings.model_dump(exclude_none=True)})
+                finalized_exporters.append(
+                    {"console": console_settings.model_dump(exclude_none=True)}
+                )
             elif exporter_name == "file":
                 file_payload = payload.copy()
                 file_payload.setdefault("path", legacy_path)
-                if "path_settings" not in file_payload and legacy_path_settings is not None:
+                if (
+                    "path_settings" not in file_payload
+                    and legacy_path_settings is not None
+                ):
                     file_payload["path_settings"] = legacy_path_settings
                 file_settings = FileExporterSettings.model_validate(file_payload)
-                finalized_exporters.append({"file": file_settings.model_dump(exclude_none=True)})
+                finalized_exporters.append(
+                    {"file": file_settings.model_dump(exclude_none=True)}
+                )
             elif exporter_name == "otlp":
                 otlp_payload = payload.copy()
                 otlp_payload.setdefault("endpoint", legacy_otlp.get("endpoint"))
                 otlp_payload.setdefault("headers", legacy_otlp.get("headers"))
                 otlp_settings = OTLPExporterSettings.model_validate(otlp_payload)
-                finalized_exporters.append({"otlp": otlp_settings.model_dump(exclude_none=True)})
+                finalized_exporters.append(
+                    {"otlp": otlp_settings.model_dump(exclude_none=True)}
+                )
             else:
                 raise ValueError(
                     f"Unsupported OpenTelemetry exporter '{exporter_name}'. Supported exporters: console, file, otlp."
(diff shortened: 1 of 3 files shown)


alienzach:
* **#551** <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/551?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/551?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/mcp_agent/config.py
```diff
@@ -581,16 +570,27 @@ class OpenTelemetrySettings(BaseModel):
     enabled: bool = False
 
     exporters: List[
-        Union[Literal["console", "file", "otlp"], OpenTelemetryExporterSettings]
+        Union[
+            Literal["console", "file", "otlp"],
+            Dict[Literal["console"], ConsoleExporterSettings | Dict],
+            Dict[Literal["file"], FileExporterSettings | Dict],
+            Dict[Literal["otlp"], OTLPExporterSettings | Dict],
```
saqadri, line 579: Why the `| Dict`?

```suggestion
            Dict[Literal["console"], ConsoleExporterSettings],
            Dict[Literal["file"], FileExporterSettings],
            Dict[Literal["otlp"], OTLPExporterSettings],
```

src/mcp_agent/config.py
```diff
@@ -599,120 +599,164 @@ class OpenTelemetrySettings(BaseModel):
     sample_rate: float = 1.0
     """Sample rate for tracing (1.0 = sample everything)"""
 
-    # Deprecated: use exporters: [{ type: "otlp", ... }]
+    # Deprecated V1 field: use exporters list with V3 syntax instead
```
saqadri: The V1 vs. V3 is quite arbitrary. Just describe what the expected syntax is

src/mcp_agent/config.py
alienzach: when we validate the schema of concrete type from yaml, we don't know which concrete type it is yet. we have to use dict to allow yaml to be deserialized into BaseModel then create concrete model after we get the type from key.

alienzach:
@saqadri PR is ready. But how do we want to approach the secret merging in this multi exporters schema?

for example, here: https://github.com/lastmile-ai/mcp-agent/blob/5e99d4dc8294fa3b5f949e2938f8ce16ec620e6e/examples/tracing/langfuse/README.md#2-set-up-secrets-and-environment-variables

we could use endpoint here to match the headers in secret and config, which mean the url have to be specified in both places as a key, also have to be unique?

or maybe I am over thinking?

examples/tracing/agent/README.md
```diff
@@ -16,10 +16,10 @@ If desired, [install Jaeger locally](https://www.jaegertracing.io/docs/2.5/getti
 otel:
   enabled: true
   exporters:
-    - type: console
-    - type: file
-    - type: otlp
-      endpoint: "http://localhost:4318/v1/traces"
+    - console: {}
+    - file: {}
```
saqadri: Maybe we set these to `file` and `console` respectively, which would have them be treated as strings

saqadri:
@alienzach I don't think we should do any sophisticated merge. The secrets one should be updated to match the schema, so you should update that readme and example to show the secrets as:


```
otel:
  enabled: true
  exporters:
    - otlp:
        endpoint: "https://us.cloud.langfuse.com/api/public/otel/v1/traces"
        headers:
          Authorization: "Basic AUTH_STRING"
```

In fact, if there are secrets to specify for the Otel exporters, that list item should just be specified in the secrets.yaml instead of config.yaml. Just validate that this case would work:

config.yaml:
```
otel:
  exporters:
    - otlp:
        endpoint: "https://us.cloud.langfuse.com/api/public/otel/v1/traces"
        headers:
          Authorization: "Basic AUTH_STRING"
```

secrets.yaml:
```
otel:
  enabled: true
  exporters:
    - otlp:
        endpoint: "https://some-other-otel-exporter"
```

Final config should include BOTH in the exporters list.

examples/tracing/llm/README.md
```diff
@@ -29,10 +29,10 @@ Then update the `mcp_agent.config.yaml` to include a typed OTLP exporter with th
 otel:
   enabled: true
   exporters:
-    - type: console
-    - type: file
-    - type: otlp
-      endpoint: "http://localhost:4318/v1/traces"
+    - console: {}
+    - file: {}
```
saqadri: For these empty cases, let's switch to the string literals...

`console`
`file`

src/mcp_agent/config.py
saqadri: Hmmm, but does it cause an issue? Why was the previous case ok where we had it as `Union[Literal["console", "file", "otlp"], OpenTelemetryExporterSettings]` (no `| Dict`)

src/mcp_agent/config.py
```diff
@@ -994,16 +1056,30 @@ def get_settings(config_path: str | None = None, set_global: bool = True) -> Set
...
-                merged[key] = deep_merge(merged[key], value)
+                merged[key] = deep_merge(merged[key], value, current_path)
+            elif (
+                key in merged
+                and isinstance(merged[key], list)
+                and isinstance(value, list)
+                and current_path == ("otel", "exporters")
+            ):
+                # Concatenate exporters lists from config and secrets
+                merged[key] = merged[key] + value
```
alienzach, line 1098: @saqadri @rholinshead call for last review before merge. especially this chunk of update, which touches the common part for merging. Test cases is in test_config_exporter.py:446-602

alienzach: This is to handle list specially in merge, instead of replace, we concat them. I currently only let this apply to otel.exporters. But I think in the future, we may need similar merge when we have secrets in list data structure again.

saqadri:
@alienzach I passed this diff through GPT-5 and it raised these two valid points:

  - Blocking: examples/workflows/workflow_evaluator_optimizer/mcp_agent.config.yaml:45 now enables both the file exporter and an OTLP exporter with a live http://
    localhost:4318 endpoint by default. As soon as someone flips enabled: true (per the README guidance) the example will try to ship spans to a collector that
    almost certainly isn’t running, which is a regression from the previous safe “console only” setup. Please restore the prior defaults (console with the OTLP
    example commented out) so enabling OTEL doesn’t break out of the box.

  - Blocking: tests/test_config_exporters.py:362 is asserting a ValueError, but the pre-validator raises a ValueError that Pydantic wraps into a ValidationError.
    The test never actually passes this with pytest.raises(ValueError) guard, so the suite will fail. Update the expectation to pytest.raises(ValidationError, …)
    (and keep the message check) to cover the real surface behavior.

saqadri:
> @alienzach I passed this diff through GPT-5 and it raised these two valid points:
> 
> * Blocking: examples/workflows/workflow_evaluator_optimizer/mcp_agent.config.yaml:45 now enables both the file exporter and an OTLP exporter with a live http://
>   localhost:4318 endpoint by default. As soon as someone flips enabled: true (per the README guidance) the example will try to ship spans to a collector that
>   almost certainly isn’t running, which is a regression from the previous safe “console only” setup. Please restore the prior defaults (console with the OTLP
>   example commented out) so enabling OTEL doesn’t break out of the box.
> * Blocking: tests/test_config_exporters.py:362 is asserting a ValueError, but the pre-validator raises a ValueError that Pydantic wraps into a ValidationError.
>   The test never actually passes this with pytest.raises(ValueError) guard, so the suite will fail. Update the expectation to pytest.raises(ValidationError, …)
>   (and keep the message check) to cover the real surface behavior.

@alienzach fixed

saqadri:
Going to merge this in so we can get this published. Nice work @alienzach!

PR #560, opened 2025-10-20
update mcp dependency; add ability to add icons

roman-van-der-krogt:
- Bump mcp to 1.18.0 (Oct 16 release)
- Add support for icons:
  - Default mcp server icon is phetch, developer can change this by setting the `icons` parameter to `MCPApp`
  - Icon for MCPApp tools (`workflows-list`, `workflows-run`, etc) is phetch. This cannot be changed by developer
  - Default tool icon (`@app.[async_]tool`) is phetch, but developer change through `icons` parameter on the decorator

## Summary by CodeRabbit

* **New Features**
  * Applications and tools can include icons; a default icon is applied when none are supplied and icons propagate to server-registered and generated workflow tools.

* **Documentation**
  * Added README section showing how to attach custom icons (local file → data URI, external URL, or icon object) with examples.

* **Tests**
  * Added tests verifying default and custom icon behavior for synchronous and asynchronous tools.

* **Examples**
  * Example workflow updated to demonstrate loading a local icon as a data URI and applying it to a tool.


roman-van-der-krogt:
Converted to draft. Earlier all tests were passing, but I rebased off the latest main and I see some test failures now

src/mcp_agent/app.py
```diff
@@ -133,6 +136,14 @@ def __init__(
         self._signal_notification = signal_notification
         self._upstream_session = upstream_session
         self._model_selector = model_selector
+        if icons:
+            self._icons = icons
+        else:
+            icon_path = Path(__file__).parent.parent.parent / "phetch.png"
```
saqadri: We should probably put the logo in `src/mcp_agent/data` (i.e. https://github.com/lastmile-ai/mcp-agent/tree/main/src/mcp_agent/data), and load it similar to how we load the benchmarks file: 

```
with (resources.files("mcp_agent.data").joinpath("icon.png").open()) as icon_file:
    icon_data = base64.standard_b64encode(icon_file.read_bytes()).decode()
```

https://github.com/lastmile-ai/mcp-agent/blob/main/src/mcp_agent/workflows/llm/llm_selector.py#L470-L473

src/mcp_agent/app.py
```diff
@@ -133,6 +136,14 @@ def __init__(
...
+            icon_data = base64.standard_b64encode(icon_path.read_bytes()).decode()
+            icon_data_uri = f"data:image/png;base64,{icon_data}"
+
+            self.icons = [Icon(src=icon_data_uri, mimeType="image/png", sizes=["48x48"])]
```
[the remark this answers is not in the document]
saqadri: This seems legit

src/mcp_agent/server/app_server.py
```diff
@@ -45,6 +47,11 @@
 _IDEMPOTENCY_KEYS_SEEN: Dict[str, Set[str]] = {}
 _IDEMPOTENCY_KEYS_LOCK = asyncio.Lock()
 
+icon_path = Path(__file__).parent.parent.parent.parent / "phetch.png"
+icon_data = base64.standard_b64encode(icon_path.read_bytes()).decode()
```
saqadri: Same as other comment

src/mcp_agent/server/app_server.py
```diff
@@ -45,6 +47,11 @@
...
+icon_data_uri = f"data:image/png;base64,{icon_data}"
```
saqadri: Does this have to be encoded or can it point to a CDN URL or S3 bucket?

src/mcp_agent/app.py
```diff
@@ -83,6 +85,7 @@ def __init__(
         signal_notification: SignalWaitCallback | None = None,
         upstream_session: Optional["ServerSession"] = None,
         model_selector: ModelSelector | None = None,
+        icons: list[Icon] | None = None,
```
saqadri, line 96: Should this be something configurable on the config as well?

saqadri (approved):
Approving to unblock but a few requests to update

src/mcp_agent/server/app_server.py
roman-van-der-krogt: It an point to a URL as well. Updated the README with an example to point that out.

saqadri: Oh awesome. In that case, should the default phetch icon be hosted on S3? That would simplify needing to encode it on each MCPApp or app_server setup

src/mcp_agent/app.py
roman-van-der-krogt: We have to include it as a parameter to the `@app.tool` decorator, so including it in the `MCPApp` creator seemed more consistent. Also, the config would work well for external images, but not so much for embedded ones. (I suppose we could detect a `file://` url and base64 encode it if we wanted to, but it still would be a bit inconsistent.) Open to changing it, though

saqadri:
@roman-van-der-krogt I think I fixed the tests too in #504 after updating the MCP dependency. Not sure if that's the same set of issues but I also had failing tests that I had to address

src/mcp_agent/server/app_server.py
roman-van-der-krogt: @saqadri ah, that's what you were after. Great suggestion. Any particular bucket? I see `publicdata.lastmileai.com`? Or should we create a new one, like `mcpac-artifacts` or something?

saqadri: publicdata is good. I think that's our existing public bucket

PR #573, opened 2025-10-23
Make Context bound to the request, allowing shared MCPApp connections

saqadri:
`app_server` was mutating the shared application `Context` with each incoming MCP request. Since Context is managed by `MCPApp`, the same `Context` instance was reused for all clients. This is fine for most things, but there are some properties in Context which are request-specific. 

Specifically, the most recent request’s `upstream_session`, identity, and logging level bled across subsequent requests. Temporal workflows and background HTTP handlers picked up whichever session happened to be set last, causing logs and notifications to go to the wrong client.

**Issues**  
- Any multi-client deployment of the MCP server: logs and callbacks could route to the wrong client once a second client connected.  
- Temporal workflow runs: the `SessionProxy` updated the global context, so unrelated workflows inherited stale sessions/identities.  
- Per-session logging level controls (`logging/setLevel`): the override was wiped at request teardown, so clients never saw the level they asked for.  
- Tests and simple scripts still ran, but only because they exercised a single client—multi-session scenarios were incorrect.

**Fix Summary**  
Replaced the implicit “mutate the shared context” behavior with an explicit request scope:

1. **Context Isolation**
   - `_enter_request_context` clones the app context for each request, remembers the parent’s session/identity, and binds the clone to a ContextVar.
   - `_exit_request_context` restores the parent’s state, clears cached “last session” markers, removes temporary attributes, and (when appropriate) leaves per-session log overrides in place until `_unregister_session`.
   - `Context.upstream_session` now returns only the request clone’s override (or an explicit app-level override), so the shared context never reports the last client’s session.

2. **Fallback and Workflow Support**
   - /internal HTTP routes use the stored execution→context mapping, but those entries are cleaned up once the request exits unless a workflow run is still active.
   - Temporal workflows still inject a `SessionProxy`; because the clone’s state is restored during teardown, the proxy no longer leaks into other runs.

3. **Logging Behavior**
   - Events inherit the request’s session id automatically; per-session log levels persist across requests until the associated workflow/session is unregistered.

4. **Tests**
   - Added isolation tests for logging, human-input flows, and temporal behavior to confirm that contexts don’t bleed between clients.

This restores the intended semantics: each request runs with its own isolated `Context`, while the shared application state remains untouched once the request finishes.

## Summary by CodeRabbit

* **New Features**
  * Request-scoped context binding to isolate logs, notifications, and sessions between concurrent clients.
  * Per-session logging level overrides for granular upstream filtering.
  * Improved session and identity propagation across request lifecycles and async boundaries.

* **Tests**
  * Comprehensive tests for request context isolation, session-scoped logging, and executor/session proxy behavior.

* **Examples**
  * Added a runnable context-isolation demo (server and clients) showing concurrent client isolation.

### DIFF
--- a/src/mcp_agent/server/app_server.py
+++ b/src/mcp_agent/server/app_server.py
@@ -130,12 +130,7 @@ async def _unregister_session(run_id: str) -> None:
             session = _RUN_SESSION_REGISTRY.pop(execution_id, None)
             _RUN_IDENTITY_REGISTRY.pop(execution_id, None)
             context_ref = _RUN_CONTEXT_REGISTRY.pop(execution_id, None)
-            session_id = _RUN_LOGGING_SESSION.pop(execution_id, None)
-            if session_id:
-                try:
-                    LoggingConfig.clear_session_min_level(session_id)
-                except Exception:
-                    pass
+            _RUN_LOGGING_SESSION.pop(execution_id, None)
             if context_ref is not None:
                 app_ref = getattr(context_ref, "app", None)
                 _clear_cached_session_refs(context_ref, session)
--- a/examples/mcp_agent_server/context_isolation/README.md
+++ b/examples/mcp_agent_server/context_isolation/README.md
@@ -3,6 +3,15 @@
 This example shows how per-request context scoping prevents logs and
 notifications from bleeding between concurrent MCP clients.
 
+## Setup
+
+- Install the example dependencies from this folder:
+  ```bash
+  uv pip install -r examples/mcp_agent_server/context_isolation/requirements.txt
+  ```
+- Optional: adjust `mcp_agent.config.yaml` if you want to tweak logging transports or
+  register additional MCP backends.
+
 ## Running the example
 
 1. Start the SSE server in one terminal:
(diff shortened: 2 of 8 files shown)


src/mcp_agent/core/context.py
```diff
@@ -105,12 +105,43 @@ class Context(MCPContext):
...
+        if request_ctx is self:
+            return getattr(self, "_upstream_session", None)
+
+        explicit = getattr(self, "_upstream_session", None)
+        if explicit is not None:
+            return explicit
+
+        parent = getattr(self, "_parent_context", None)
+        if parent is not None:
+            return getattr(parent, "_upstream_session", None)
```
roman-van-der-krogt, line 145: Should we return `parent.upstream_session()` here?  It's not likely that we're going to have sessions nested several layers deep, but just in case

src/mcp_agent/server/app_server.py
```diff
@@ -60,14 +66,16 @@
 
 
 if TYPE_CHECKING:
-    from mcp_agent.core.context import Context
+    pass
```
roman-van-der-krogt: Might as well remove the `if` altogether

src/mcp_agent/server/app_server.py
```diff
@@ -79,17 +87,63 @@
...
+        return
+    if session is not None:
+        try:
+            if getattr(target, "_last_known_upstream_session", None) is session:
+                setattr(target, "_last_known_upstream_session", None)
+        except Exception:
+            pass
+    if session_id is not None:
+        try:
+            if getattr(target, "_last_request_session_id", None) == session_id:
```
roman-van-der-krogt: What is the point of the `_last_request_session_id`? I see the `_last_known_upstream_session` is being returned in `_get_fallback_upstream_session`, but `last_request_session_id` doesn't seem to be used anywhere? Should `get_fallback_upstream_session` also return the `_last_request_session_id`?

PR #576, opened 2025-10-23
Update cloud example requirements to specify 'mcp-agent[cli]'

StreetLamb:
`uv run mcp-agent login` fails if user runs example without installing mcp-agent optional dependencies prior.

## Summary by CodeRabbit

* **Chores**
  * Updated mcp-agent package dependencies to include CLI extras across example projects. This configuration update has been applied consistently to chatgpt_app, hello_world, mcp, and temporal examples. The changes ensure unified dependency specifications and enhanced CLI functionality support across all example implementations.


saqadri:
Good catch. "uv run mcp-agent" works, but I guess otherwise it wouldn't work?

StreetLamb:
`uv run mcp-agent` doesn't work for me either if I remove .venv and install the example without cli extra.

PR #583, opened 2025-10-23
Fix coroutine 'MCPAppClient.get_app_by_name' was never awaited

StreetLamb:
`uvx mcp-agent cloud servers describe my-agent` fails as coroutine 'MCPAppClient.get_app_by_name' was never awaited

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an async operation issue in the app ID retrieval function to ensure proper coroutine handling.


petersonbill64 (approved):
thanks - I think I fixed this locally but failed to push it to the PR in time

PR #587, opened 2025-10-23
Fix/template updates

rholinshead:
# Summary
Fix the server template for init and exclude factory and token for now until we can make sure they work (and have correct README contents) --> or just remove them in a later PR.

## Testing
- Installed, ran, deployed & ran the server template
- Made sure minimal template still works

## Summary by CodeRabbit

* **New Features**
  * requirements.txt is now automatically created when generating templates.

* **Documentation**
  * Updated and reorganized template README files for improved guidance.

* **Changes**
  * Server entry script renamed to main.py.
  * Token and factory scaffolding templates are now disabled.
  * Updated default LLM provider configuration.
  * Modified server startup execution method.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,5 +13,8 @@
       "mcp-agent.secrets.yaml",
       "mcp_agent.secrets.yaml"
     ]
+  },
+  "files.watcherExclude": {
+    "**/target": true
   }
 }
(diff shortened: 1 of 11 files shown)


src/mcp_agent/cli/commands/init.py
```diff
@@ -153,8 +153,8 @@ def init(
     scaffolding_templates = {
         "basic": "Simple agent with filesystem and fetch capabilities",
         "server": "MCP server with workflow and parallel agents",
-        "token": "Token counting example with monitoring",
-        "factory": "Agent factory with router-based selection",
+        # "token": "Token counting example with monitoring",
```
rholinshead, line 156: Just removing these for now; generally not sure about their usefulness as templates (vs just examples in the repo) but also they were copying unrelated README so would at least need an updated README on top of validating that they work (e.g. we know token counting doesn't work for cloud workflow runs atm)

PR #594, opened 2025-10-27
Remove server hostname function and simplify server naming

jtcorbett:
### TL;DR

Simplified server naming and added a restart notification for MCP client installations.

### What changed?

- Removed the `_server_hostname()` function that generated server names from URLs
- Replaced dynamic server naming with a simple default value of "mcp_agent"
- Added a notification message informing users they may need to restart their MCP client after installation
- Removed redundant app info fetching in the ChatGPT client flow

## Summary by CodeRabbit

- **New Features**
  - Added a reminder after successful installation prompting users to restart the MCP client for changes to take effect.

- **Refactor**
  - Simplified server name selection during installation; when no name is provided, it now defaults to the application name or "mcp_agent", reducing hostname derivation complexity.

### DIFF
--- a/src/mcp_agent/cli/commands/install.py
+++ b/src/mcp_agent/cli/commands/install.py
@@ -39,6 +39,7 @@ from mcp_agent.cli.core.constants import (
     ENV_API_BASE_URL,
     ENV_API_KEY,
 )
+from mcp_agent.cli.core.utils import run_async
 from mcp_agent.cli.exceptions import CLIError
 from mcp_agent.cli.mcp_app.api_client import MCPAppClient
 from mcp_agent.cli.utils.ux import (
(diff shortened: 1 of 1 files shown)


jtcorbett:
* **#594** <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/594?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lastmile-ai/mcp-agent/594?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/mcp_agent/cli/commands/install.py
```diff
@@ -417,7 +382,7 @@ def install(
         )
         return
 
-    server_name = name or _server_hostname(server_url, app_name)
+    server_name = name or "mcp_agent"
```
saqadri: We should use the same logic as `deploy` for name definition. I believe that uses "default". That said, I'm also ok with "mcp_agent"

jtcorbett: this will get the app_name from the server (which is set by `deploy`), so only in really strange cases where that never gets set will this be used

PR #601, opened 2025-10-31
Allow env to be saved as secrets, add new app-scoped secrets command and materialize an mcp_agent.deployed.config.yaml

saqadri:
# Summary

TLDR -- tries to implement these ergonomics: 
image

Declaring environment variables meant editing raw YAML, secrets were processed server-side, and there was no convenient CLI to inspect or rotate env secrets after initial deploys. This PR brings the developer experience closer to Vercel’s `env` workflow by:

- letting projects declare deployment env vars directly in `mcp_agent.config.yaml`
- materializing those envs into secrets during `cloud deploy`, alongside a frozen config snapshot (`mcp_agent.deployed.config.yaml`)
- exposing `mcp-agent cloud secrets` commands (list/add/remove/pull) so teams can manage env secrets without redeploying

# Functional Changes

## Config schema & runtime

- `Settings.env` now accepts a list of strings or single-key dicts, e.g.:

  ```yaml
  env:
    - OPENAI_API_KEY                         # resolved via os.environ
    - {SUPABASE_URL: "https://db.example"}    # fallback literal
  ```

- Added validation helpers plus `iter_env_specs()` to preserve ordering and union string/dict entries.
- `MCPApp` injects literal fallbacks into `os.environ` if the process is missing the variable; handles (`mcpac_sc_*`) are ignored so the control plane can still supply real values.

## Deployment workflow

- New `materialize_deployment_artifacts()` helper invoked from `cloud deploy`:
  - writes `mcp_agent.deployed.config.yaml` (user config snapshot)
  - reads each `env` spec, resolves values (`os.environ` → fallback literal), creates/updates secrets (`apps/<app_id>/env/<KEY>`), and writes their handles under `env:` in `mcp_agent.deployed.secrets.yaml`.
  - existing handles are reused to keep deployments idempotent.
- Deploy logs now call out the new artefacts.

## CLI enhancements

- `mcp-agent cloud secrets` command group:
  - `list [app]` – show env secrets with masked handles.
  - `add <key> <value> [app]` – create or update a secret value.
  - `remove <key> [app]` – delete the secret handle.
  - `pull [app] --output mcp_agent.cloud.secrets.yaml` – download resolved values without touching `mcp_agent.secrets.yaml`.
- Commands accept `--config-dir`, `--api-url`, and `--api-key` and fall back to project defaults when possible.

## Documentation

- Updated CLI reference, cloud quickstart, and manage-secrets guides with:
  - `env` declaration examples and how they feed into `mcp_agent.deployed.secrets.yaml`.
  - the presence and purpose of `mcp_agent.deployed.config.yaml`.
  - detailed walkthrough for the new `cloud secrets` subcommands.
  - callouts mirroring Vercel’s `env pull` experience.

## Backend coordination

- Bundle processor can now consume the pre-materialized config (`BUNDLE_CONFIG_FILENAME`) and still applies cloud overrides (Temporal, SSE). It skips gracefully if the file is missing or invalid.

# Usage Walkthrough

```bash
uvx mcp-agent init --template basic cloud-demo
cd cloud-demo
export OPENAI_API_KEY="sk-..."               # values resolved at deploy time
export SUPABASE_URL="https://db.example.com"

# Declare env vars in mcp_agent.config.yaml:
# env:
#   - OPENAI_API_KEY
#   - {SUPABASE_URL: "https://db.example.com"}  # optional fallback

uvx mcp-agent deploy cloud-demo --non-interactive

uvx mcp-agent cloud secrets list cloud-demo
uvx mcp-agent cloud secrets add EXTRA_TOKEN "super-secret" cloud-demo
uvx mcp-agent cloud secrets pull cloud-demo
uvx mcp-agent cloud secrets remove EXTRA_TOKEN cloud-demo
```

`mcp_agent.deployed.secrets.yaml` now contains:

```yaml
env:
  - OPENAI_API_KEY: mcpac_sc_...
  - SUPABASE_URL: mcpac_sc_...
```

And `mcp_agent.deployed.config.yaml` mirrors the project config after overrides, letting the control plane skip redundant transformations.

## Test Plan

Here's the comprehensive test plan I used to validate the change:

### 1. Author env declarations

1. Open `mcp_agent.config.yaml`.
2. Add an `env` list with mixed syntax:
   ```yaml
   env:
     - OPENAI_API_KEY              # no fallback, must come from env
     - {SUPABASE_URL: "https://db.example.com"}  # fallback literal
   ```
3. Ensure `OPENAI_API_KEY` is set in your shell; leave `SUPABASE_URL`
   unset to exercise the fallback prompt later.
4. Validate schema support by running `uvx mcp-agent build --dry-run` (or
   letting your editor’s YAML validation pass).

Expected: No schema errors, config lints cleanly.

---

### 2. Deploy and capture env secrets

1. Run `uvx mcp-agent deploy <app-name> --non-interactive` in the project
   directory.
2. Observe prompts/logs:
   - Deployment should read `OPENAI_API_KEY` from the environment.
   - For `SUPABASE_URL`, deployment should auto-use the fallback literal.
3. After success, inspect generated files:
   - `mcp_agent.deployed.secrets.yaml` contains:
     ```yaml
     env:
       - OPENAI_API_KEY: mcpac_sc_...
       - SUPABASE_URL: mcpac_sc_...
     ```
   - `mcp_agent.deployed.config.yaml` matches the resolved config but has no
     raw secret values (search for `sk-` or `SUPABASE_URL` literal; only env
     placeholder should remain).

Expected: Deployment succeeds; both deployed files exist; config snapshot
contains no plaintext secrets.

---

### 3. Verify runtime env injection

1. Tail logs via `uvx mcp-agent cloud logger tail <app-name> --since 5m`.
2. Push a request to the server (e.g., via the MCP client or sample workflow)
   that prints/uses the environment variables.

Expected: App has access to `OPENAI_API_KEY` and `SUPABASE_URL` via `os.environ`
without checking in secrets locally.

---

### 4.1 List
`uvx mcp-agent cloud secrets list <app-name>`

Expected: Table shows both env keys with masked handles.

#### 4.2 Add
`uvx mcp-agent cloud secrets add EXTRA_TOKEN super-secret <app-name>`
`uvx mcp-agent cloud secrets add --from-env-file .env.local --app <app-name>`

Expected: Command prints “Created secret for EXTRA_TOKEN …”.

#### 4.3 Pull
`uvx mcp-agent cloud secrets pull <app-name> --format env`

Expected:
- `.env.mcp-cloud` is created (or updated when `--force` is provided) with `KEY=value` pairs, quoted when necessary.
- Running `uvx mcp-agent ...` commands afterward automatically picks up `.env.mcp-cloud` (after any local `.env`) with no manual sourcing.
- Re-run with `--format yaml --output pulled.yaml` to confirm the YAML path still works when needed.

#### 4.4 Remove & redeploy
1. `uvx mcp-agent cloud secrets remove EXTRA_TOKEN <app-name>`
2. Confirm `list` no longer shows `EXTRA_TOKEN`.
3. Run `uvx mcp-agent deploy <app-name> --non-interactive` again.

Expected: Deploy succeeds even though the previous handle was deleted (CLI
provisions a new handle automatically). `mcp_agent.deployed.secrets.yaml`
contains whichever env keys remain.

---

### 5. Semantics of fallbacks/prompts

1. Unset `SUPABASE_URL` and remove the fallback from `mcp_agent.config.yaml`.
2. Run `uvx mcp-agent deploy <app-name>` (interactive mode).

Expected: CLI prompts “Enter value for environment variable 'SUPABASE_URL'…”.
Entering a value stores it as a secret; redeploying should reuse the handle.
3. Modify `.env.mcp-cloud`, re-run a local command (e.g., `uvx mcp-agent build --dry-run`), and confirm the new value is picked up without editing `mcp_agent.secrets.yaml`, proving the `.env` loading order works (`.env` > `.env.mcp-cloud` > config fallbacks).

---

### 6. Bundle integrity

1. Inspect the deployment bundle (optional) to confirm it contains:
   - `mcp_agent.deployed.config.yaml`
   - `mcp_agent.deployed.secrets.yaml` (handles)
   - *Not* `mcp_agent.secrets.yaml`
2. Confirm `.mcpacignore` or custom ignore files behave as documented.

---

## Summary by CodeRabbit

* **New Features**
  * Declare env vars in config to capture them at deploy; deployment materializes config and secrets files for reproducible bundles
  * New cloud secrets commands: list, add, remove, pull for post-deploy inspection and rotation

* **Documentation**
  * Updated deployment quickstart and CLI reference with env-var capture, materialized artifacts, and secrets management workflows

* **Tests**
  * Added tests covering env spec parsing and deployment artifact materialization (including secret handle reuse and fallbacks)


docs/cloud/mcp-agent-cloud/manage-secrets.mdx
```diff
@@ -110,6 +129,24 @@ Example output:
 
 Subsequent runs reuse stored values; use `--dry-run` to validate without persisting changes.
 
+### Manage environment secrets after deployment
+
+Use the `cloud secrets` commands to inspect or rotate values captured via the `env` list:
+
+```bash
+uvx mcp-agent cloud secrets list my-app
```
rholinshead, line 138: Is `my-app` the name?

docs/cloud/mcp-agent-cloud/manage-secrets.mdx
```diff
@@ -52,6 +53,24 @@ notion:
...
+  - {SUPABASE_URL: "https://db.example.com"}  # fallback literal if the env var is unset
+```
+
+Each entry can be either a plain string (read from `os.environ[KEY]`) or a single-key mapping that supplies a fallback value when the environment variable is missing. During `mcp-agent deploy` the CLI:
+
+1. Resolves the value (environment first, then fallback).
+2. Stores it as a deployment secret named `apps/<app_id>/env/<KEY>`.
+3. Inserts the resulting handle into the `env:` list inside `mcp_agent.deployed.secrets.yaml`.
+
+The runtime still references the value via `os.environ[KEY]`; the cloud control plane materialises the secret before your app starts.
```
rholinshead, line 73: We'll need to do the materialization part in the deployment backend still, right?

docs/cloud/mcp-agent-cloud/manage-secrets.mdx
```diff
@@ -110,6 +129,24 @@ Example output:
...
+uvx mcp-agent cloud secrets list my-app
+uvx mcp-agent cloud secrets add SUPABASE_URL https://db.example.com my-app
+uvx mcp-agent cloud secrets remove SUPABASE_URL my-app
+uvx mcp-agent cloud secrets pull my-app --output mcp_agent.cloud.secrets.yaml
+```
+
+- `list` masks secret handles so you can confirm which keys are present.
+- `add` updates the stored value (under the hood it calls the Secrets API and rewrites the handle if needed).
+- `remove` deletes the value entirely.
+- `pull` resolves values to a local YAML file (`mcp_agent.cloud.secrets.yaml` by default) without touching `mcp_agent.secrets.yaml`, mirroring Vercel’s `env pull`.
```
rholinshead: What do you do with the `cloud.secrets` yaml afterwards? Wondering if it's a bit too much complexity having yet another yaml file vs just printing to the console or something

docs/reference/cli.mdx
```diff
@@ -109,6 +110,21 @@ uvx mcp-agent deploy research-app \
 
 Use `--dry-run` for validation without uploading, and let the CLI guide you through classifying developer versus user secrets.
 
+When deployment completes you will see two new artefacts next to your source:
```
rholinshead, line 113:
```suggestion
When deployment completes you will see two new artifacts next to your source:
```

docs/reference/cli.mdx
```diff
@@ -198,6 +214,12 @@ This workflow lets you adopt a published server template, supply your credential
   - `--secrets-output-file/-o` chooses where prompted secrets are written (defaults to `mcp_agent.configured.secrets.yaml`).  
   - `--params` lists required secrets and exits.  
   - `--dry-run` validates using mock clients without persisting anything.  
+
+- **Secrets (`mcp-agent cloud secrets …`)** – manage environment secrets captured during deploy.  
+  - `secrets list [APP]` displays stored environment keys with masked handles.
```
rholinshead, line 219: Just noting the possibility for confusion with these commands being named 'secrets' but actually specifically being _env_ secrets and not general secrets that may be used/stored in the secrets config; would something like `env`, `env_vars` or `env_secrets` be a clearer command name?

docs/cloud/mcp-agent-cloud/manage-secrets.mdx
rholinshead: Oh, actually I think below with the env handling from the config in mcp-agent it should already be handled because the handle will already be resolved as normal in the deployed secrets file and then the env specs are put back into env by mcp-agent itself

src/mcp_agent/cli/cloud/commands/deploy/materialize.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    config_path.parent.mkdir(parents=True, exist_ok=True)
+    materialized = settings.model_dump(
+        mode="python",
+        exclude_none=True,
+        exclude_unset=True,
+        exclude_defaults=not include_overrides,
+    )
+    with open(config_path, "w", encoding="utf-8") as handle:
+        yaml.safe_dump(materialized, handle, default_flow_style=False, sort_keys=False)
```
rholinshead: Can you double-check this works with ouath examples? For context, when trying to deploy oauth/auth config settings to cloud on Friday I was seeing an error resulting from the `AnyHttpUrl` types in the config, which can't serialize properly. I didn't have enough time to fully fix in the deployment daemon (will continue on that tomorrow) - but it would either require custom dump handling or use mode="json". The latter adds an extra "/" at the end, though. But in any case, my understanding is safe_dump will actually raise an exception instead, so apps with oauth wouldn't be able to be deployed without error during materialization

rholinshead: Side-note, fwiw claude thinks mode="json" would be safe _except for_ specific provider handling:
```
Based on this detailed analysis, here's what I found:

  🎯 The Good News: Most Critical Paths Are Safe

  ✅ Cache Keys - All Canonicalized

  - Resource metadata cache: URLs go through normalize_resource() first
  - Authorization metadata cache: URLs go through _canonicalize_url() first
  - Verdict: Trailing slashes won't cause cache issues

  ✅ Token Storage - Indirectly Canonicalized

  - Resource URLs: Processed by normalize_resource() before storage
  - Authorization server URLs: Processed by _canonicalize_url() before storage
  - Verdict: Trailing slashes won't cause token storage mismatches

  ✅ URL Comparisons - Both Sides Canonicalized

  - Authorization server validation canonicalizes both the provided URL and the resolved issuer
  - Verdict: Trailing slashes won't cause comparison failures

  ⚠️ The Potential Issues

  1. OAuth Provider API Calls

  # These are used RAW from metadata/config:
  authorize_url = httpx.URL(str(auth_metadata.authorization_endpoint) + "?" + params)
  token_endpoint = str(auth_metadata.token_endpoint)
  redirect_uri = redirect_uri  # From config

  Impact: If OAuth providers do strict string matching on redirect URIs, a mismatch between registered https://example.com/callback and Pydantic-normalized
  https://example.com/callback/ could cause failures.

  2. Callback URL Registration

  When you register your OAuth app with a provider, you specify exact callback URLs. If:
  - You register: https://example.com/callback
  - But Pydantic normalizes to: https://example.com/callback/
  - And that gets sent to the OAuth provider

  Then the OAuth provider might reject it due to URL mismatch.

  🎯 Conclusion

  The trailing slash issue is likely NOT a problem for most functionality because:

  1. Internal caching and storage is safe - everything gets canonicalized
  2. URL comparisons are safe - both sides get canonicalized
  3. Most URL construction is safe - rstrip('/') handles both cases

  The only real risk is OAuth provider callback URL validation, which depends on:
  - Whether OAuth providers do exact string matching on redirect URIs
  - Whether the registered callback URLs match what gets sent after Pydantic normalization

  For most OAuth providers, this probably isn't an issue since they typically normalize URLs themselves. But it's worth testing with actual OAuth flows to be sure.
```

src/mcp_agent/cli/cloud/commands/deploy/materialize.py
```diff
@@ -0,0 +1,264 @@
...
+
+    env_specs = _normalize_env_specs(settings)
+
+    # Always materialize the user config into a deployed variant
+    deployed_config_path = config_dir / MCP_DEPLOYED_CONFIG_FILENAME
+    _persist_deployed_config(deployed_config_path, settings)
+
+    if not env_specs:
+        # Nothing further to do; ensure secrets file exists if previously created
+        if not deployed_secrets_path.exists():
```
rholinshead, line 208: Can you clarify this case? It seems to just be creating an empty deployed secrets file if it doesn't exist yet. But for apps that don't have secrets, we wouldn't want that, right?

src/mcp_agent/cli/cloud/commands/deploy/materialize.py
```diff
@@ -0,0 +1,192 @@
...
+        if handle:
+            run_async(secrets_client.set_secret_value(handle, value))
+        else:
+            handle = run_async(
+                secrets_client.create_secret(
+                    name=secret_name,
+                    secret_type=SecretType.DEVELOPER,
+                    value=value,
+                )
+            )
```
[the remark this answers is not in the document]
rholinshead: Probably legit

src/mcp_agent/cli/cloud/commands/deploy/materialize.py
```diff
@@ -0,0 +1,264 @@
...
+            added_path = True
+
+        if module_name in sys.modules:
+            del sys.modules[module_name]
+
+        module = importlib.import_module(module_name)
+        module_file = Path(getattr(module, "__file__", "")).resolve()
+        if not module_file or project_root not in module_file.parents:
+            typer.secho(
+                f"Module 'main' resolved outside project directory ({module_file}); skipping MCPApp load.",
```
rholinshead, line 321: Is this sys/module manipulation an additional security over checking main.py exists?

docs/cloud/mcp-agent-cloud/manage-secrets.mdx
rholinshead: Ah, ok, looks like it's any of the identifiers (name,id,server)

src/mcp_agent/cli/cloud/commands/secrets/main.py
```diff
@@ -0,0 +1,311 @@
...
+    )
+
+
+def _env_secret_prefix(app_id: str) -> str:
+    return f"apps/{app_id}/env/"
+
+
+def _load_existing_handles(client: SecretsClient, app_id: str) -> Dict[str, str]:
+    prefix = _env_secret_prefix(app_id)
+    secrets = run_async(client.list_secrets(name_filter=prefix))
```
rholinshead, line 133: (later) We can also add the association between the app and secrets in the DB to be able to query more efficiently

rholinshead (commented):
Overall looks good to me but I have a strong suspicion the config materialization is going to surface exceptions for urls in the config, so we should resolve that before landing

src/mcp_agent/cli/cloud/commands/deploy/materialize.py
rholinshead: Ok, https://github.com/lastmile-ai/mcp-agent/pull/605 should be all we need to allow us to just use `mode="json"` here when yaml dumping

saqadri:
Addressed several PR comments. Going to merge this in to make forward progress, and will address all remaining PR comments in a followup.

docs/cloud/mcp-agent-cloud/manage-secrets.mdx
saqadri: Yep, it can be any app identifier (url, id, name)

docs/cloud/mcp-agent-cloud/manage-secrets.mdx
saqadri: Exactly... it should leverage the existing approach and work fully.

docs/cloud/mcp-agent-cloud/manage-secrets.mdx
saqadri: Agreed. Removed the cloud.secrets.yaml in favor of materializing a `.env.mcp-cloud` instead. Also updated MCPApp to load_dotenv

docs/reference/cli.mdx
saqadri: Great point, will reconsider. Maybe I should call it `env`

src/mcp_agent/cli/cloud/commands/deploy/materialize.py
saqadri: Thanks, resolved this and also added a test

tests/app/test_dotenv_loading.py
```diff
@@ -0,0 +1,45 @@
...
+    monkeypatch.delenv("MY_SECRET", raising=False)
+
+
+def test_config_fallback_overrides_existing_env(monkeypatch):
+    monkeypatch.setenv("SUPABASE_URL", "original")
+    settings = Settings(env=[{"SUPABASE_URL": "https://fallback.example"}])
+    app = MCPApp(settings=settings)
+    app._apply_environment_bindings()
+
+    assert os.environ["SUPABASE_URL"] == "https://fallback.example"
```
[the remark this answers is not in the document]
saqadri: disagree, the config's env settings should take precedence

Issue #538, opened 2025-10-02
`uv run mcp-agent init` runs slow, would be good to have some indicator of progress


andrew-lastmile:
Takes 5-10 seconds for `uv run mcp-agent init` to show

```
Initializing MCP-Agent project
Directory: /Users/andrew_lastmile_ai/Documents/GitHub/mcp-agent-testing
Template: basic - Simple agent with filesystem and fetch capabilities
```

Because there's no indicator that something is happening, I'm conflicted on whether to just cancel the execution.

jaydaftari:
Hi, Have created a Pr for this #608, Please review my changes.

rholinshead:
Completed by #608

PR #608, opened 2025-11-07
feature for: uv run mcp-agent init runs slow, would be good to have s…

jaydaftari:
Solution for : uv run mcp-agent init runs slow, would be good to have some indicator of progress #538.
Added another file: `main_bootstrap.py` because the imports were taking a long time, and I displayed an indicator until all the imports were loaded. Also updated the entry point in the pyproject.toml file corresponding to that.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced CLI startup experience with loading indicator during initialization through optimized bootstrap process.

* **Chores**
  * Minor formatting and whitespace adjustments across configuration and example files.


saqadri:
This is great! Thank you for adding this @jaydaftari. There is a small thing I would prefer, specifically to only show the spinner when `Console(stderr=True).is_terminal` is true so that we disable the spinner in non-interactive environments (e.g. running the CLI in CI/scripts. etc.) But we can do that as a fast-follow.

jaydaftari:
Hi @saqadri kindly, review the raised PR for this:  https://github.com/lastmile-ai/mcp-agent/pull/613

Issue #596, opened 2025-10-28
PLEASE add LMStudio support


javi2375:
Thank you for this awesome project!
I would love to use it and I know plenty of people are using local models now to build agents. LMStudio is the most accessible platform, present in all OSs and hardware configurations available (mac, intel, amd, cuda). I know you have ollama support, but nowadays it is fallen in favor of other clients for local models due to their issues implementing newer llama.cpp runtimes.

I would love to be able to use your project with LMstudio and so would many other people who are not able to run vllm or trying to get away from cloud providers. Please consider it!! 

Humbly requested from a non-coder physician who wants to improve healthcare with agents. 

Thanks in advance!

rholinshead:
Hey @javi2375, I've added a basic AugmentedLLM for LM Studio in https://github.com/lastmile-ai/mcp-agent/pull/622, so it will be available with our next published package. In the meantime, if you have a local mcp-agent repo cloned you can test it out with that (example in https://github.com/lastmile-ai/mcp-agent/tree/main/examples/lm_studio)

PR #622, opened 2025-12-02
Add AugmentedLLM for LM Studio

rholinshead:
## Summary
As requested in #596, this PR adds basic support for LM Studio models by leveraging the [OpenAI Compatibility Endpoints](https://lmstudio.ai/docs/developer/openai-compat#openai-compatibility-endpoints) supported by LM Studio, extending the AugmentedLLMOpenAI to use the endpoints with overridden base url pointing to local LM Studio server.

A couple notes:
- The LM Studio api does not support tool calling combined with structured output so we handle `generate_structured` as 2 calls --> one to generate with tool calling, one to generate structured with the response model
- We override the model selection to prioritize the default model from the config instead of from the benchmarks since there are no LM Studio model benchmarks, and it makes more sense to use the provided config value anyway

## Testing
```
make format
make lint
make tests
```

Added example, which works for both default model `openai/gpt-oss-20b` and `deepseek/deepseek-r1-distill-qwen-14b`

## Summary by CodeRabbit

* **New Features**
  * Added LM Studio integration to run local models and produce structured outputs via the agent framework.
  * Added a sample interactive agent demo and structured-output example for local model usage.

* **Documentation**
  * Added comprehensive example README, configuration guide, and requirements for LM Studio usage.

* **Refactor**
  * Improved provider config handling for OpenAI-backed flows.

* **Tests**
  * Added tests covering LM Studio integration, config injection, and model selection logic.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/examples/lm_studio/mcp_agent.config.yaml
+++ b/examples/lm_studio/mcp_agent.config.yaml
@@ -14,7 +14,7 @@ mcp:
   servers:
     filesystem:
       command: "npx"
-      args: ["-y", "@modelcontextprotocol/server-filesystem"]
+      args: ["-y", "@modelcontextprotocol/server-filesystem", "."]
 
 lm_studio:
   # base_url defaults to http://localhost:1234/v1
--- a/examples/lm_studio/mcp_agent.config.yaml
+++ b/examples/lm_studio/mcp_agent.config.yaml
@@ -3,7 +3,7 @@ $schema: ../../schema/mcp-agent.config.schema.json
 execution_engine: asyncio
 logger:
   transports: [console, file]
-  level: debug
+  level: info
   progress_display: true
   path_settings:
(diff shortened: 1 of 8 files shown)


examples/lm_studio/mcp_agent.config.yaml
```diff
@@ -0,0 +1,21 @@
+$schema: ../../schema/mcp-agent.config.schema.json
+
+execution_engine: asyncio
+logger:
+  transports: [console, file]
+  level: debug
```
StreetLamb: the console is very noisy and I cant see the expected output shown in the README file. Should we set it to `level: info` instead?

StreetLamb (approved):

examples/lm_studio/mcp_agent.config.yaml
rholinshead: Done

PR #623, opened 2025-12-05
Fix breaking type change in mcp 1.20.0 URL elicitation

petersonbill64:
Support MCP `ElicitRequestParams` as a union type by subclassing its two member types instead of the union itself.

## Summary by CodeRabbit

* **Chores**
  * Bumped project version to 0.2.6.
  * Updated MCP dependency requirement from >=1.19.0 to >=1.20.0.
  * Internal improvements to elicitation parameter handling and type safety for enhanced robustness.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/mcp_agent/server/app_server.py
+++ b/src/mcp_agent/server/app_server.py
@@ -1023,7 +1023,8 @@ def create_mcp_server_for_app(app: MCPApp, **kwargs: Any) -> FastMCP:
                 CreateMessageRequestParams,
                 CreateMessageResult,
                 ElicitRequest,
-                ElicitRequestParams,
+                ElicitRequestFormParams,
+                ElicitRequestURLParams,
                 ElicitResult,
                 ListRootsRequest,
                 ListRootsResult,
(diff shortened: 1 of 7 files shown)


src/mcp_agent/workflows/llm/augmented_llm.py
```diff
@@ -31,6 +31,7 @@
     SamplingMessage,
     TextContent,
     PromptMessage,
+    Tool,  # noqa: F401 - Required to resolve forward reference in CreateMessageRequestParams
```
saqadri, line 34: I don't follow this one

src/mcp_agent/server/app_server.py
```diff
@@ -1164,7 +1171,7 @@ class AuthToken(BaseModel):
             callback_future = await callback_registry.create_handle(flow_id)
             req = ElicitRequest(
                 method="elicitation/create",
-                params=ElicitRequestParams(
+                params=ElicitRequestFormParams(
```
saqadri, line 1174: @romanvanderkrogt should the auth flow now use url based elicitation?

src/mcp_agent/workflows/llm/augmented_llm.py
petersonbill64: I get
```
pydantic.errors.PydanticUserError: `RequestParams` is not fully defined; you should define `Tool`, then call `RequestParams.model_rebuild()`
```
when running the elicitation example without this

PR #634, opened 2026-01-15
Add real-time streaming support for Anthropic and Bedrock

MarianaGiraldoL:
This PR adds comprehensive real-time streaming capabilities to mcp-agent, enabling LLM responses to be displayed incrementally as they're generated (similar to ChatGPT). Users can now stream text deltas, monitor tool executions in real-time, and build responsive UIs with live progress updates.

Fixes: #13, #307

### Core Streaming API
- `generate_stream()` - Full-featured streaming with structured events
- `generate_str_stream()` - Simplified text-only streaming convenience method
- **Event-driven architecture** with 9 structured event types
- **Real-time tool execution visibility** during multi-iteration agentic loops
- **Per-iteration and total token usage** reporting

### Provider Support
✅ Anthropic - Native streaming via client.messages.stream()
✅ Bedrock - Native streaming via converse_stream()
⏳ OpenAI, Azure, Google (planned for future PRs)

### Event Types
```python
StreamEventType.TEXT_DELTA        # Incremental text content
StreamEventType.THINKING          # Extended thinking (supported models)
StreamEventType.TOOL_USE_START    # Tool call initiated
StreamEventType.TOOL_RESULT       # Tool execution result
StreamEventType.TOOL_USE_END      # Tool call completed
StreamEventType.ITERATION_START   # Agentic iteration begins
StreamEventType.ITERATION_END     # Iteration complete (with usage)
StreamEventType.COMPLETE          # Generation finished (total usage)
StreamEventType.ERROR             # Error occurred
```

### Architecture

- **BaseModel:** New `StreamEvent` Pydantic model with full validation
- **Decorator Enhancement:** Extended `@track_tokens()` to support async generators
- **History Management:** Preserved across streaming calls with `use_history` parameter
- **Error Handling:** Comprehensive try-catch with structured ERROR events

### Token Tracking

- Per-iteration usage reported in `ITERATION_END` events
- Total usage reported in `COMPLETE` event (both providers)
- Automatic accumulation via enhanced `@track_tokens()` decorator
- Compatible with existing token counter infrastructure

## 🔄 Backward Compatibility

**100% Backward Compatible** - No breaking changes:

- Existing generate() method unchanged
- Streaming is opt-in via new methods
- All existing code continues to work
- No API changes to non-streaming methods

## Summary by CodeRabbit

* **New Features**
  * Real-time LLM streaming with structured StreamEvent updates and a text-only convenience stream (generate_stream / generate_str_stream)
[...]

PR #648, opened 2026-03-18
fix: skip reasoning_effort for gpt-5.4+ with function tools

dragon1086:
## Problem

gpt-5.4 and gpt-5.4-mini (released March 17, 2026) reject `reasoning_effort` when function tools are present on `/v1/chat/completions`:

```
Error code: 400 - "Function tools with reasoning_effort are not supported 
for gpt-5.4 in /v1/chat/completions. Please use /v1/responses instead."
```

Older models (gpt-5.2, gpt-5.3, o1, o3, o4) don't have this restriction.

## Fix

In `generate()`, skip adding `reasoning_effort` to the request when the model starts with `gpt-5.4` and tools are present. That's it — one conditional.

`generate_structured()` doesn't use function tools, so no change needed there.

I verified this against the actual OpenAI API — gpt-5.2 with tools + reasoning_effort still works, gpt-5.4 without tools still works, only the specific combo of gpt-5.4 + tools + reasoning_effort is broken.

## Tests

3 new tests covering the fix + backward compat. All 27 tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added a public streaming API surface (placeholder) for future streaming support.

* **Refactor**
  * Adjusted when reasoning_effort is included based on model variant and tool usage; structured outputs now always include reasoning_effort for reasoning models.

* **Tests**
  * Added tests validating reasoning_effort propagation, token-field behavior, tool/no-tool scenarios, streaming paths, and related error handling.

### DIFF
--- a/src/mcp_agent/workflows/deep_orchestrator/orchestrator.py
+++ b/src/mcp_agent/workflows/deep_orchestrator/orchestrator.py
@@ -795,3 +795,12 @@ class DeepOrchestrator(AugmentedLLM[MessageParamT, MessageT]):
             response_model=response_model,
             request_params=RequestParams(max_iterations=1),
         )
+
+    async def generate_stream(
+        self,
+        message,
+        request_params=None,
+    ):
+        """Streaming is not yet implemented for DeepOrchestrator."""
+        raise NotImplementedError("Streaming not yet implemented for DeepOrchestrator")
+        yield  # Make this a generator function
--- a/src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
+++ b/src/mcp_agent/workflows/evaluator_optimizer/evaluator_optimizer.py
@@ -1,6 +1,6 @@
 import contextlib
 from enum import Enum
-from typing import Callable, List, Optional, Type, TYPE_CHECKING
+from typing import AsyncIterator, Callable, List, Optional, Type, TYPE_CHECKING
 from pydantic import BaseModel, Field
 
 from mcp_agent.tracing.semconv import GEN_AI_AGENT_NAME
@@ -15,6 +15,7 @@ from mcp_agent.workflows.llm.augmented_llm import (
 )
 from mcp_agent.agents.agent import Agent
 from mcp_agent.logging.logger import get_logger
+from mcp_agent.workflows.llm.streaming_events import StreamEvent
 
 if TYPE_CHECKING:
     from mcp_agent.core.context import Context
@@ -430,6 +431,15 @@ class EvaluatorOptimizerLLM(AugmentedLLM[MessageParamT, MessageT]):
                     span.set_attribute("unstructured_response", response_str)
 
             return res
+
+    async def generate_stream(
+        self,
+        message,
+        request_params=None,
+    ) -> AsyncIterator[StreamEvent]:
+        """Streaming is not yet implemented for this provider."""
+        raise NotImplementedError("Streaming not yet implemented for this provider")
+        yield  # Make this a generator function
 
     def _build_eval_prompt(
         self, original_request: str, current_response: str, iteration: int
--- a/src/mcp_agent/workflows/llm/augmented_llm_azure.py
+++ b/src/mcp_agent/workflows/llm/augmented_llm_azure.py
@@ -1,7 +1,7 @@
 import asyncio
 import functools
 import json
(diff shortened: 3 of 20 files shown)


dragon1086:
Note: the CI failures are pre-existing on main — all caused by the recently added `generate_stream` abstract method that several LLM subclasses (Google, LMStudio, Ollama, Swarm, ParallelLLM, etc.) haven't implemented yet.

Our `test_augmented_llm_openai.py` tests all pass (27/27), including the 3 new ones added in this PR. We also added a `_TestableOpenAIAugmentedLLM` subclass to work around the same issue in the OpenAI tests.

dragon1086:
## Note on additional commits (generate_stream stubs)

The last two commits (`b8bdb05d`, `c2290b4f`) are slightly outside the original scope of this PR, but I had to add them to unblock my production use case.

**Background**: PR #634 ("Add real-time streaming support for Anthropic and Bedrock") added `generate_stream` as an `@abstractmethod` on the base `AugmentedLLM` class, but only implemented it for `AnthropicAugmentedLLM` and `BedrockAugmentedLLM`. This means **every other subclass** that doesn't implement `generate_stream` will fail to instantiate:

```
TypeError: Can't instantiate abstract class OpenAIAugmentedLLM with abstract method generate_stream
TypeError: Can't instantiate abstract class EvaluatorOptimizerLLM with abstract method generate_stream
```

I hit both of these in production, so I added minimal stubs (raise `NotImplementedError`) for `OpenAIAugmentedLLM` and `EvaluatorOptimizerLLM` to satisfy the abstract contract.

The remaining subclasses (`Orchestrator`, `DeepOrchestrator`, `ParallelLLM`, `Swarm`, `LLMRouter`, `GoogleAugmentedLLM`, `AzureAugmentedLLM`) have the same issue but are outside my scope. This should probably be addressed upstream in a follow-up to #634.

dragon1086:
## Fix: `generate_stream` abstract method not implemented in all subclasses

CI was failing with 78 test failures + 77 errors due to `TypeError: Can't instantiate abstract class ... with abstract method generate_stream`.

### Root Cause
PR #634 added `generate_stream` as an `@abstractmethod` to `AugmentedLLM` base class, but only implemented it in `AnthropicAugmentedLLM`, `OpenAIAugmentedLLM`, and `BedrockAugmentedLLM`. All other subclasses (both source and test mocks) were missing the implementation, causing instantiation failures.

### Changes
Added `NotImplementedError` stubs (same pattern as the existing `EvaluatorOptimizerLLM` stub) to:

**Source classes (7):**
- `AzureAugmentedLLM`
- `GoogleAugmentedLLM`
- `ParallelLLM`
- `LLMRouter`
- `Orchestrator`
- `DeepOrchestrator`
- `Swarm`

**Test mock classes (8):**
- `DummyLLM`, `MockAugmentedLLM` (×3), `MockAugmentedLLMWithTokens` (×3), `_DummyLLM`, `_MockLLM`

### Result
- Before: 78 failed, 77 errors
- After: 1503 passed, 1 pre-existing failure (`test_bedrock_streaming` — unrelated token counting assertion)

dragon1086:
## Additional fix: Bedrock streaming test mock

Also fixed the remaining `test_single_turn_text_streaming` failure which was pre-existing from #634.

**Root cause:** The mock `create_mock_stream_response` was missing the `metadata` event that Bedrock sends after `messageStop`. The real Bedrock API sends usage data in a `{"metadata": {"usage": {...}}}` event at the end of the stream, but the mock only included it at the top-level response dict — which `generate_stream` doesn't read (it reads from the stream events).

**Fix:** Added `{"metadata": {"usage": usage}}` event to the mock stream, matching Bedrock's actual behavior.

**Result:** All 1505 tests passing, 4 skipped, 0 failures.

dragon1086:
## Added GPT-5.5 compatibility (430b9e4d)

The existing check `model.startswith("gpt-5.4")` did not cover `gpt-5.5` models.

**Change:**
- `model.startswith("gpt-5.4")` → `re.match(r"gpt-5\.[4-9]", model)`
- Now covers gpt-5.4 through gpt-5.9, including gpt-5.5 and any future gpt-5.x minor versions

GPT-5.5 shares the same limitation as GPT-5.4: `reasoning_effort` is not supported alongside function tools on `/v1/chat/completions` (returns 400). The regex approach future-proofs this for upcoming gpt-5.x releases as well.

PR #665, opened 2026-04-05
ci: add HOL skill-publish validate workflow

internet-dot:
I opened this as a small validate-only check for the skill metadata in the repo.

A couple of repo-specific details I checked first:
- Build effective agents using Model Context Protocol and simple workflow patterns
- Build effective agents with Model Context Protocol using simple, composable patterns
- Full MCP support: It _fully_ implements MCP, and handles the pesky business of managing the lifecycle of MCP server connections so you don't have to

This adds one workflow for `.` and leaves the runtime code alone.
It only runs the schema and trust checks for the skill metadata in validate mode.

If you would rather place the workflow under a different filename or point it at a different skill directory, I can adjust the branch.


internet-dot:
Hi 👋 - these are intentional design choices:

1. **SKILL.md/skill.json don't exist yet** - Correct! This workflow runs when you ADD these files. It's a pre-flight check for skill publishing.

2. **id-token: write** - Standard GitHub OIDC. Validates job identity but grants no access to secrets. Validate mode is secretless.

3. **Path filters** - The workflow triggers on changes to these files wherever they are in the repo.

Let me know if you'd like to adjust the workflow!

PR #675, opened 2026-04-21
fix(agent): skip failed capability lookups for unavailable servers

### LINKED ISSUE #671: get_capabilities_task silently returns exception objects in ServerCapabilities dict under partial server failure
## Bug Description

`get_capabilities_task` uses `asyncio.gather(..., return_exceptions=True)` to fetch capabilities from all connected MCP servers, then zips the results directly into a `Dict[str, ServerCapabilities]` without checking whether any result is an exception. When a server is unavailable or raises, the exception object ends up as a value in the dict, silently violating the return type.

## Location

`src/mcp_agent/agents/agent.py:1541–1545`

```python
capabilities: List[ServerCapabilities] = await asyncio.gather(
    *[aggregator.get_capabilities(server_name=n) for n in server_names],
    return_exceptions=True,
)
server_capabilities = dict(zip(server_names, capabilities))  # exceptions mixed in
```

## Reproduction

1. Configure an agent with two MCP servers, one of which is unreachable.
2. Call `get_capabilities_task` for that agent.
3. The returned dict contains an `Exception` instance as the value for the failing server instead of `ServerCapabilities`.
4. Any downstream code accessing `.tools`, `.prompts`, or `.resources` on that value raises `AttributeError`.

## Impact

When any server in a multi-server agent setup is temporarily unavailable, the capabilities call appears to succeed (no exception is raised) but returns a corrupt dict. Downstream consumers crash with confusing attribute errors rather than a clear "server X is unavailable" message. The failure mode is especially hard to diagnose because it is silent at the point of collection.

## Suggested Fix

Check each result after gathering and exclude failures explicitly:

```python
results = await asyncio.gather(
    *[aggregator.get_capabilities(server_name=n) for n in server_names],
    return_exceptions=True,
)
server_capabilities = {}
for name, result in zip(server_names, results):
    if isinstance(result, BaseException):
        logger.warning("Failed to get capabilities for %s: %s", name, result)
    else:
        server_capabilities[name] = result
```

---
Found via automated codebase analysis. Happy to submit a PR if this is confirmed.

MukundaKatta:
## Summary
- stop mixing exception objects into the get_capabilities_task result map when one server fails
- log and skip failed capability lookups instead of returning a corrupt Dict[str, ServerCapabilities]
- add a regression test covering partial failure across multiple servers

Closes #671

## Verification
- python3 -m pytest tests/agents/test_agent_tasks_concurrency.py -k get_capabilities_skips_failed_servers (not runnable here because the sandbox Python environment is missing the mcp package required by tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * When fetching capabilities from multiple servers, failed or missing responses are now skipped instead of appearing in results.
  * Added warning logs that identify which servers failed capability retrieval and why, improving troubleshooting visibility.

* **Tests**
  * Added tests verifying failed or None capability responses are omitted from aggregated results.

### DIFF
--- a/src/mcp_agent/agents/agent.py
+++ b/src/mcp_agent/agents/agent.py
@@ -1538,11 +1538,19 @@ class AgentTasks:
             if not server_name:
                 # If no server name is provided, get capabilities for all servers
                 server_names: List[str] = aggregator.server_names
-                capabilities: List[ServerCapabilities] = await asyncio.gather(
+                capabilities = await asyncio.gather(
                     *[aggregator.get_capabilities(server_name=n) for n in server_names],
                     return_exceptions=True,
                 )
-                server_capabilities = dict(zip(server_names, capabilities))
+                for name, capability in zip(server_names, capabilities):
+                    if isinstance(capability, BaseException):
+                        logger.warning(
+                            "Failed to get capabilities for server '%s': %s",
+                            name,
+                            capability,
+                        )
+                        continue
+                    server_capabilities[name] = capability
             else:
                 # If a server name is provided, get capabilities for that server
                 server_capabilities[server_name] = await aggregator.get_capabilities(
--- a/src/mcp_agent/agents/agent.py
+++ b/src/mcp_agent/agents/agent.py
@@ -1550,6 +1550,12 @@ class AgentTasks:
                             capability,
                         )
                         continue
+                    if capability is None:
+                        logger.warning(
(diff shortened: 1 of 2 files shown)


MukundaKatta:
Addressed @coderabbitai review: added explicit `capability is None` skip branch before the assignment (with a matching 'no capabilities returned' warning), and extended the concurrency tests to cover that path via a `return_none_for` set on `FakeAggregator`.

PR #707, opened 2026-06-24
fix(openai): send tool results as string content

### LINKED ISSUE #25: content in OpenAI ChatCompletion messages is a list, not string

asdaxzcwqsa:
## Summary
Fix OpenAI tool-response formatting so tool messages are sent back as string content instead of content-part lists.

## Problem
When a tool returns `TextContent`, `execute_tool_call()` was converting the result into a `tool` message with `content=[...]`. OpenAI ChatCompletion expects tool message content to be a string, which can trigger validation errors on the next completion call.

## Fix
- route tool results through `OpenAIConverter.convert_tool_result_to_openai()`
- flatten any additional follow-up messages returned by the converter
- add a regression test that asserts the second OpenAI request contains a tool message with string content
- make the OpenAI LLM test fixture concrete so the regression test file runs again

## Verification
- `uv sync --extra openai --dev`
- `uv run pytest tests/workflows/llm/test_augmented_llm_openai.py -q`

Fixes #25

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ordering of tool-related messages after tool execution to keep follow-up content consistent.
  * Enhanced handling of tool call results when tools return multiple messages, ensuring the full sequence is sent back correctly.
  * Fixed tool-result message formatting so tool content is always sent as plain text (including mixed text/image cases) with matching tool-call IDs.
* **Tests**
  * Extended tool-call mocking and added coverage for multi-tool result formatting and message payload expectations.

### DIFF
--- a/tests/workflows/llm/test_augmented_llm_openai.py
+++ b/tests/workflows/llm/test_augmented_llm_openai.py
@@ -11,7 +11,7 @@ from openai.types.chat import (
 )
 from pydantic import BaseModel
 
-from mcp.types import TextContent, SamplingMessage, PromptMessage
+from mcp.types import CallToolResult, ImageContent, TextContent, SamplingMessage, PromptMessage
 
 from mcp_agent.config import OpenAISettings
 from mcp_agent.workflows.llm.augmented_llm_openai import (
@@ -111,6 +111,40 @@ class TestOpenAIAugmentedLLM:
                         "arguments": json.dumps(tool_args),
                     },
                 )
+            ],
+        )
+        choice = Choice(
+            finish_reason=finish_reason,
+            index=0,
+            message=message,
+        )
+        return ChatCompletion(
+            id="chatcmpl-123",
+            choices=[choice],
+            created=1677858242,
+            model="gpt-4o",
+            object="chat.completion",
+            usage=usage,
+        )
+
+    @staticmethod
+    def create_multi_tool_use_response(tool_calls, finish_reason="tool_calls", usage=None):
+        """
+        Creates a tool use response with multiple tool calls for testing.
+        """
+        message = ChatCompletionMessage(
+            role="assistant",
+            content=None,
+            tool_calls=[
+                ChatCompletionMessageToolCall(
+                    id=tool_call["id"],
+                    type="function",
+                    function={
+                        "name": tool_call["name"],
+                        "arguments": json.dumps(tool_call["arguments"]),
+                    },
+                )
+                for tool_call in tool_calls
             ],
         )
         choice = Choice(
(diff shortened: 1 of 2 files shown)


asdaxzcwqsa:
Addressed the review feedback by keeping all tool-role messages grouped before any follow-up messages from mixed-content tool results, and extended the regression test to cover the multi-tool tuple branch as well. Running uv run pytest tests/workflows/llm/test_augmented_llm_openai.py -q still passes locally.

tests/workflows/llm/test_augmented_llm_openai.py
```diff
@@ -363,6 +402,88 @@ async def custom_side_effect(*args, **kwargs):
...
+
+        assert first_tool_message["tool_call_id"] == "tool_123"
+        assert first_tool_message["content"] == "First tool result"
+        assert not isinstance(first_tool_message["content"], list)
+
+        assert second_tool_message["tool_call_id"] == "tool_456"
+        assert second_tool_message["content"] == "Second tool result"
+        assert not isinstance(second_tool_message["content"], list)
+
+        assert follow_up_message["tool_call_id"] == "tool_456"
```
sviat-barbutsa, line 484: Hi! One thing I'd adjust in the mixed-content path: this assertion locks in `tool_call_id` on a `role: "user"` follow-up message, but OpenAI chat message params only use `tool_call_id` on `role: "tool"` messages. So the next provider call can still contain a message that does not match the OpenAI schema.

Could you drop `tool_call_id` from the synthetic user message and update the regression test to assert that only the tool message carries it? The ordering test can still keep the follow-up message right after all tool responses!

PR #728, opened 2026-08-08
fix(workflow): fail Temporal run on @app.workflow_run exceptions

### LINKED ISSUE #503: @app.workflow_run does not fail the run on Exception
When an Exception is raised within an `@app.workflow_run`-wrapped function, it results in a `Workflow Task Failed` in temporal but the run itself stays in 'Running' state. For example:

```py
@app.workflow
class PauseResumeWorkflow(Workflow[str]):
    """
    A workflow that demonstrates Temporal's signaling capabilities.
    This workflow pauses execution and waits for a signal before continuing.
    """

    @app.workflow_run
    async def run(
        self, message: str = "This workflow demonstrates pause and resume functionality"
    ) -> WorkflowResult[str]:
        """
        Run the pause-resume workflow.

        Args:
            message: A message to include in the workflow result.

        Returns:
            WorkflowResult containing the processed data.
        """
        print(f"Starting PauseResumeWorkflow with message: {message}")
        print(f"Workflow is pausing, workflow_id: {self.id}, run_id: {self.run_id}")
        print(
            "To resume this workflow, use the 'workflows-resume' tool or the Temporal UI"
        )

        raise RuntimeError("TEST")
```

arimu1:
## Summary

Fixes #503: exceptions raised inside `@app.workflow_run` (e.g. `RuntimeError`) were treated by Temporal as **workflow task** failures only, so the Temporal run stayed in `Running` while the task retried. Raising a non-retryable `ApplicationError` already failed the run correctly.

This change converts plain exceptions from the workflow `run` path into a non-retryable `WorkflowApplicationError` so the **run** fails and the error is surfaced.

## Changes

- Add `reraise_with_workflow_failure()` in `mcp_agent.executor.errors`:
  - Converts plain exceptions → non-retryable `WorkflowApplicationError`
  - Preserves existing `WorkflowApplicationError` / Temporal `FailureError` (including `ApplicationError` with user-set `non_retryable`)
  - Best-effort: sets workflow `status` to `"error"` and records the original error on `state`
- Use it in `@app.workflow_run` and in auto-generated workflow `_run` (tool-style workflows)

## Test plan

- [x] `tests/executor/test_errors.py` — conversion + preserve ApplicationError/FailureError
- [x] `tests/test_app.py` — `@app.workflow_run` converts `RuntimeError("TEST")` (issue repro) and preserves explicit ApplicationError
- [x] Existing workflow_run decorator tests still pass

```bash
uv sync --extra temporal --group dev
uv run pytest tests/executor/test_errors.py tests/test_app.py::TestMCPApp::test_workflow_run_converts_runtime_error tests/test_app.py::TestMCPApp::test_workflow_run_preserves_application_error tests/test_app.py::TestMCPApp::test_workflow_run_decorator_default -q
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow error handling across supported execution modes.
  * Unexpected failures now preserve their original type and message during local execution.
  * Temporal workflows now report unexpected failures as non-retryable application errors.
  * Existing classified workflow and application errors retain their original details.
  * Workflow status and error information are recorded consistently when initialization or execution fails.


src/mcp_agent/app.py
```diff
@@ -677,40 +677,54 @@ def workflow_run(self, fn: Callable[..., R], **kwargs) -> Callable[..., R]:
...
+                    workflow_cls, f"_decorated_{method_name}", None
+                )
+                if class_decorated:
+                    return await class_decorated(*args, **kwargs)
 
-            # Fall back to the original function
-            return await fn(*args, **kwargs)
+                # Fall back to the original function
+                return await fn(*args, **kwargs)
+            except Exception as e:
```
sviat-barbutsa, line 725: Nice fix for the Temporal path. I think this conversion should stay limited to Temporal, though.

`workflow_run` is used by both `asyncio` and `temporal`, so this also replaces ordinary asyncio/domain exceptions with `WorkflowApplicationError`. That changes the behavior for callers that catch the original exception type, even though asyncio does not have the Temporal retry issue. The new test uses `execution_engine = "asyncio"`, so it currently locks in that cross-engine change.

Could you, please, preserve the original exception for non-Temporal execution and add a regression test for it, while keeping this conversion for Temporal?

arimu1: Thanks for the review — agreed. Updated in 46203f8: `reraise_with_workflow_failure` now takes `for_temporal`, and `@app.workflow_run` passes `for_temporal=(execution_engine == "temporal")`, so asyncio re-raises the original exception while Temporal still gets `WorkflowApplicationError`. Added `test_workflow_run_preserves_runtime_error_for_asyncio` plus explicit Temporal conversion coverage.

PR #734, opened 2026-08-17
fix(logging): respect configured log level inside Temporal workflows

### LINKED ISSUE #382: Log level doesn't seem to be respected in temporal worker?
> It's hard to look for this particular message: `ValueError: Server aggregrator for agent 'finder' not found`.
> 
> It looks like when temporal executor is used, it doesn't respect the `loggersettings`, I have it set to console + error but it still logs debug logs. 

 _Originally posted by @kawolum in [#380](https://github.com/lastmile-ai/mcp-agent/issues/380#issuecomment-3198488708)_

Log level doesn't seem to be respected in temporal worker?

```
settings = Settings(
    execution_engine="temporal",
    logger=LoggerSettings(type="console", level="error")
...
```

When I set this, the worker seems to still log debug logs, can you confirm if `LoggerSettings` is expected to work for Temporal?

tonycoder-hub:
Fixes #382.

## Problem
`Logger._emit_event` has a Temporal workflow branch that never reaches `AsyncEventBus.emit`. Events go through the session proxy, `mcp_forward_log`, or `emit_with_stderr_transport`, none of which apply `logger.level`. A worker set to `level="error"` still prints debug.

## Fix
`LoggingConfig.allows_event` at the top of that branch. Consults local and upstream filters so `logging/setLevel` still works. No-op when logging was never configured.

## Test plan
- [x] New below-level test fails on current main (debug still printed) and passes after
- [x] `uv run pytest tests/logging/test_temporal_workflow_logging.py -q` → 2 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow event forwarding now respects separate local and upstream logging filters.
  * Events not allowed by either filter are suppressed.
  * Events permitted only for local logging are no longer forwarded upstream.
  * Stderr fallback output now follows the local logging filter.
* **Tests**
  * Added coverage for severity filtering, local output, and upstream forwarding behavior.


tonycoder-hub:

PR #645, opened 2026-03-10
fix: improve streaming support for Bedrock

MarianaGiraldoL:
## Summary
This PR contains additional fixes for streaming support following the previous feature/streaming-support branch changes.

## Changes
- **augmented_llm.py**: Remove `@abstractmethod` decorator from `generate_stream` method to allow default implementations
- **augmented_llm_bedrock.py**: 
  - Fix tool input parsing to handle empty strings (tools with no arguments)
  - Create deep copy of tool use block dict to avoid mutation issues
  - Restructure tool result collection to batch all tool results into a single message
- **test_bedrock_streaming.py**: Add metadata events to test cases for completeness

## Test plan
- [x] Run existing tests
- [x] Verify streaming with tool calls works correctly
- [x] Test tools with empty arguments

## Summary by CodeRabbit

* **New Features**
  * Tool invocation parsing improved and tool results are now sent as aggregated messages in both streaming and non‑streaming flows.

* **Bug Fixes**
  * Empty or missing tool inputs are handled gracefully; tool-result collection and consolidation are more robust.

* **Tests**
  * Added metadata/usage event helper and extended tests to validate aggregated tool-result payloads and metadata inclusion.

* **Refactor**
  * Streaming support for LLM integrations made optional for implementations.

PR #735, opened 2026-08-17
fix: stop treating AugmentedLLM.generate_stream as abstract

tonycoder-hub:
## Summary
`AugmentedLLM.generate_stream` is `@abstractmethod` despite having a default body, so every subclass that does not implement streaming cannot be instantiated. Main Checks have been red since #634.

Drop the decorator and make the fallback an async generator so iterating it raises `NotImplementedError` instead of `TypeError`. That also unblocks `generate_str_stream`.

Open #645 (last updated March) also removes the decorator as part of a larger Bedrock streaming rework and leaves the fallback as a coroutine. This is the focused suite-unblocking slice only.

## Test plan
- [x] New fallback tests error on current main and pass after
- [x] `uv run pytest tests -q` → 1502 passed, 4 skipped (was 97 failed / 107 errors)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling for language model integrations that do not support streaming.
  - These integrations can now be used without implementing streaming, with a clear error provided when streaming is requested.
  - Preserved existing error propagation behavior for streamed responses.
  - Improved streaming test coverage, including token usage metadata in Bedrock responses.


tonycoder-hub:

PR #736, opened 2026-08-17
fix(app): bind each global workflow task to its own bound method

tonycoder-hub:
## Summary
`MCPApp._register_global_workflow_tasks()` builds a `_bound_adapter` closure inside a loop. The adapter reads `func` and `self_ref` from the enclosing scope, so after the loop every registered activity runs the last bound method on the last instance.

Move adapter construction into `_make_bound_task_adapter(func, self_ref)` so each task captures its own function and instance. Distinct from #734/#735 and from #729 (activity timeouts).

## Test plan
- [x] New `test_register_global_workflow_tasks_keeps_bound_method_identity` fails on current main and passes after
- [x] `uv run pytest tests/test_app.py -q` → 39 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed registration of globally registered bound workflow tasks so they retain the correct instance and function identity.
  * Ensured activity metadata and task execution results remain accurate after registration.

* **Tests**
  * Added regression coverage for bound workflow-task registration and registry state restoration.


tonycoder-hub:

PR #737, opened 2026-08-17
fix(mcp): stop discarding explicit server_name in MCPAggregator.call_tool

tonycoder-hub:
call_tool re-declared server_name = None before the if server_name guard, so the explicit-server path was dead. Tests: uv run pytest tests/mcp/ -q -> 40 passed. Distinct from #734/#735/#736.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool routing when an explicit server name is provided.
  * Preserved tool names unchanged, including names containing separators.
  * Ensured requests are sent to the specified persistent server.

* **Tests**
  * Added coverage for explicit server-name routing and separator-containing names.


tonycoder-hub:

PR #738, opened 2026-08-17
fix(workflows): join attributed source strings in FanIn.aggregate_message_lists

tonycoder-hub:
The loop builds source_message_strings with Source N prefixes, then joins the raw source_messages. That drops attribution and TypeErrors on non-string MessageT inputs. Tests: uv run pytest tests/workflows/parallel/test_fan_in.py -q -> 27 passed. Distinct from #734/#735/#736/#737.

## Summary by CodeRabbit

- **Bug Fixes**
  - Aggregated workflow responses now preserve source attribution for each message.
  - Non-text messages, including structured data and numeric values, are rendered correctly in aggregated results.

- **Tests**
  - Added coverage to verify source attribution and formatting across supported message types.


tonycoder-hub:

PR #739, opened 2026-08-17
fix(logging): do not stringify repeated objects in JSONSerializer

tonycoder-hub:
_processed_objects accumulated every visited id and never popped on unwind, so a second visit to the same interned int/bool or shared dict was replaced by str(obj). serialize({a: 1, b: 1}) became {a: 1, b: "1"}. Genuine cycles still stringify.

Also deferred the logger import so the module can be imported without a cycle.

Tests: 6 new tests in tests/logging/test_json_serializer.py passed. Distinct from #734-#738.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON serialization of repeated values and nested objects.
  * Prevented infinite recursion when serializing self-referencing dictionaries and lists.
  * Improved serializer reliability across repeated operations.
  * Ensured the logging serializer can be imported independently.


tonycoder-hub:

PR #740, opened 2026-08-17
fix(mcp): treat ServerRegistry read_timeout_seconds as seconds, not days

tonycoder-hub:
ServerRegistry.start_server built the session read timeout with timedelta(config.read_timeout_seconds). timedelta first positional arg is days, so a configured 30-second timeout became 30 days. MCPConnectionManager and the http/sse timeout lines in the same function already use timedelta(seconds=...).

Tests: uv run pytest tests/mcp/ -q — 40 passed. Distinct from #734-#739.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of configured server read timeouts for more reliable timeout behavior.
  - Preserved the expected behavior when no timeout is configured.

- **Tests**
  - Added coverage for configured and omitted read-timeout settings.

PR #741, opened 2026-08-17
fix(executor): forward positional args in AsyncioExecutor.execute_many

tonycoder-hub:
AsyncioExecutor.execute_many accepts *args and documents them as positional arguments to pass to each task, but dropped them. _execute_task then caught the TypeError and returned it as the task result, so callers got a list of TypeError objects instead of a crash. execute, execute_streaming, and TemporalExecutor.execute_many already forward *args.

Tests: uv run pytest tests/executor/test_asyncio_executor.py -q — 3 passed; tests/executor — 122 passed. Distinct from #734-#740.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed asynchronous batch execution so positional arguments are correctly passed to every task.
  * Ensured consistent argument handling across standard, batch, and streaming execution modes.

* **Tests**
  * Added coverage validating positional and keyword argument forwarding.


tonycoder-hub:

PR #742, opened 2026-08-17
fix(executor): await async signal_notification callbacks

tonycoder-hub:
Executor.wait_for_signal called signal_notification without awaiting it. SignalWaitCallback is an async def, so the coroutine was created and discarded (RuntimeWarning: never awaited). MCPApp(signal_notification=...) dashboards never fired.

Await the return value only when it is a coroutine, so the documented sync lambda still works.

Tests: uv run pytest tests/executor tests/test_app.py -q — 159 passed. Distinct from #734-#741.

## Summary by CodeRabbit

* **Bug Fixes**
  * Signal notifications now work correctly with both synchronous and asynchronous callbacks.
  * Callback results and notification metadata are consistently delivered.

* **Tests**
  * Added coverage for synchronous and asynchronous signal-notification handling.


tonycoder-hub:

PR #743, opened 2026-08-17
fix(cli): treat /sse/ and /mcp/ as transport suffixes

tonycoder-hub:
parse_server_url matched the transport suffix against the raw path, so "/sse/".endswith("/sse") was False. https://example.com/sse/ became https://example.com/sse/mcp and was labeled http. Same for /mcp/ -> /mcp/mcp. configure.py already strips slashes in its fallback.

Strip trailing slashes before matching; the returned URL stays as the user typed it.

Tests: uv run pytest tests/cli/utils/test_url_parser.py -q — 6 passed. tests/cli: 208 passed, 4 skipped. Distinct from #734-#742.


tonycoder-hub:

